PR #28, opened 2021-04-21
Remove excessive copyright

BassmentChemist:


effgarces:
Thanks for the contribution, there was already changes to the copyright messages in another PR, so after merging those, these are no longer current.

Issue #33, opened 2021-04-27
smarty php7 count modifier


bpinsard:
Emails sent by booked contained the following errors:
```
Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/booked/tpl_c/1d000b594d0f33f56aa31ad4351fb635d53c04fc_0.file.ReservationCreatedAdmin.tpl.php on line 127

Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/booked/tpl_c/1d000b594d0f33f56aa31ad4351fb635d53c04fc_0.file.ReservationCreatedAdmin.tpl.php on line 146

Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/booked/tpl_c/1d000b594d0f33f56aa31ad4351fb635d53c04fc_0.file.ReservationCreatedAdmin.tpl.php on line 161

Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/booked/tpl_c/1d000b594d0f33f56aa31ad4351fb635d53c04fc_0.file.ReservationCreatedAdmin.tpl.php on line 180
```
Caused by passiong something else than an array.

It looks like it can be solved by https://github.com/smarty-php/smarty/issues/459
replacing all `|count` by `|default:array()|count`

effgarces:
Thanks for letting us know and for the pull request I will be taking a look  at this in a couple of minutes.

PR #34, opened 2021-04-27
replace all `|count` by `|default:array()|count`

bpinsard:
Should solve #33.
It's a bit a brutal search/replace (some of which might not be necessary if the array passed to count is never null.


effgarces:
There are some issues in several files (I haven't check all 128) with encoding, for instance lang/ee_ee/ReservationCreated.tpl  were `Väljak` became `V�ljak`
I was checking why there  was so many conflicts and it seems that your repo is a bit out of date. There have been some changes in an effort to make development easier, as such the develop branch will have all of latest code and the master the final versions.

bpinsard:
Rebased on develop, seems to fix the merge.

effgarces:
Yes, seems perfect.

PR #36, opened 2021-04-27
Company Name and Url option added in config

BassmentChemist:
Added option in config to add company name and URL. Shows in footer, unless left blank. Cleaned up some formatting


effgarces:
I will definitely cherry pick some of your commits later today, thanks for the contribution.

PR #37, opened 2021-04-28
revert modifier fix, make all php 7.4 compatible

bpinsard:
I reverted the modifier changes and did again a search/replace of all the instance of implodes not compatible with php7.4. This should fix notably the schedule where nothing was displayed.


effgarces:
The two commits have been cherry picked on to develop so solve  a couple of conflicts.

PR #124, opened 2022-06-21
Feature - Admin-only Check in/out

dharmsworth:
New configuration options to restrict check-in/out to administrators only. We are using this for equipment loaning where staff need to confirm collection and return.


effgarces:
Nice work, I will try to test this in the coming days/weeks. There's already a plugin that provides the functionality of admin only check-in/out but having it built-in is a lot better

Issue #151, opened 2022-11-19
Reporting not working anymore


zed76:
my hosting provider switched to PHP 8.1, now Reporting is not working anymore.

The errors log reads 

`Stack trace:
#0 /is/htdocs/wp13484192_NWVLTH8V9S/www/test/lib/Application/Reporting/ReportDefinition.php(4): require_once()
#1 /is/htdocs/wp13484192_NWVLTH8V9S/www/test/lib/Application/Reporting/namespace.php(7): require_once('/is/htdocs/wp13...')
#2 /is/htdocs/wp13484192_NWVLTH8V9S/www/test/Presenters/Reports/CommonReportsPresenter.php(6): require_once('/is/htdocs/wp13...')
#3 /is/htdocs/wp13484192_NWVLTH8V9S/www/test/Pages/Reports/CommonReportsPage.php(5): require_once('/is/htdocs/wp13...')
#4 /is/htdocs/wp13484192_NWVLTH8V9S/www/test/Web/reports/common-reports.php(4): require_once('/is/htdocs/wp13...')
#5 {main} [File=/is/htdocs/wp13484192_NWVLTH8V9S/www/test/lib/Common/Logging/ExceptionHandler.php,Line=19]
2022-11-19T10:44:50+01:00 [32133] ERROR default - [User=Admin Admin (admin@example.com)] Uncaught exception: Error: Class "ReportCell" not found in /is/htdocs/wp13484192_NWVLTH8V9S/www/test/lib/Application/Reporting/ReportColumn.php:5`

Any ideas how to fix this or is this a problem specific to my hosting provider (Hosteurope)? I double checked this error by doing a fresh install.

effgarces:
The problem is with librebooking and PHP8.1, not your hosting provider, I believe that it's the last thing still causing issues with php8, at the moment I still haven't managed to fix it and will probably only have time to dive again on to the code in a few weeks

zed76:
thank you a lot for your effort to keep librebooking running. I will wait for a fix then 😊

Issue #146, opened 2022-10-14
issue when running with PHP version 8.0, isue is NOT present with php verison 7.4 nor 8.1 . Can't find compatibilty info in cpanel for Boooked.


ldillard79:
![booked issue with phpv8 pic2](https://user-images.githubusercontent.com/89486778/195956787-73dc87d3-2f96-409d-80ef-b14293aacfb0.JPG)
![booked issue with phpv8 pic1](https://user-images.githubusercontent.com/89486778/195956789-9bff373b-ae00-4854-b0e1-e25b6988ad41.JPG)
![booked issue with phpv8 pic3](https://user-images.githubusercontent.com/89486778/195956791-0df54f20-edb7-4aa9-bc61-f7a7ef463f9a.JPG)

effgarces:
All these warnings can be safely ignored, and in 7.x PHP they were not present, they started showing up in php 8, and are also present in 8.1, what probably is happening is that on the 8.1 cpanel config warnings have been disabled.
I'm currently working on "fixing" all of these".

PR #160, opened 2022-12-17
PHP8 fixes

### LINKED ISSUE #58: PHP 8 Support
It seems that the latest version does not support PHP 8, as it does not use namespaces in PHP, redeclaring system classnames instead (e.g. Attribute).

If anyone could solve this that would be great, otherwise, if I have time, I might look at it, but I'm not familiar with the codebase..

Yeicor:
This adds lots of fixes for PHP8 warnings and deprecations, without losing compatibility with PHP7.

This also includes a fix for the reporting feature that wasn't working.

All changes are tested under PHP 8.2.1RC1.

The last commit also modifies the docker-compose file, mounting the app directory for instantaneous updates during development.

Closes #58, #146, #151...


effgarces:
Wow, one of the best christmas gifts so far, thanks so much for the contribution.

PR #235, opened 2023-09-23
Issue 166

colisee:
### Changes
- Fix the `ldap` plugin that was broken with the introduction of php-8.0 and php-8.1

### Remarks:
- One day, we should remove the plugins that come with external php packages like Net_LDAP2


effgarces:
Thanks for the contribution, as soon as I can I will make a new release.

sdh91:
Any update to this? I appear to be having the same issue, however, in 2.8.6.2 it appears this was merged already. So I'm not sure if there is more to this or if there is something else wrong.

PR #271, opened 2023-11-22
Update INSTALLATION.md

F0r3n5iC-Sw33t13:
Change the log files instructions. It was set to the old way but it is changed in the config file


F0r3n5iC-Sw33t13:
Also changed the composer layout. It isn't needed anymore when you set it up

F0r3n5iC-Sw33t13:
Remove the Gitter box, it isn't online anymore

F0r3n5iC-Sw33t13:
Added a whole translation series in Dutch

PR #276, opened 2023-11-29
Update README.md

F0r3n5iC-Sw33t13:
Added some dates in README.md by the version numbers

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,16 +39,16 @@ Please consult the wiki for more help <https://github.com/LibreBooking/app/wiki>
 
 ## Release Notes
 
-#### 2.8.6.1
+#### 2.8.6.1 - 26th of September 2023
 Mainly Bug fixes, special mention for the ldap plugin, more details at <https://github.com/LibreBooking/app/commits/develop>
 
-#### 2.8.6
+#### 2.8.6 - 18th of April 2023
 
 Librebooking now has PHP8 support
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -57,15 +57,15 @@ Fork this repo, contribute and help keep it alive
 Small update to fix a security issue
 
 
-#### 2.8.5.4
+#### 2.8.5.4 - 3th of September 2021
```
JohnVillalovos, line 60: `3th` is not a thing.  `3rd` would be correct.

But personally I would do dates like: 2021-09-03 or 03-Sep-2021

effgarces: Agree, in fact I would argue for using 2021-09-03, as it's probably the most universal version

F0r3n5iC-Sw33t13: Agreed, I will change it!

colisee:
Hi @F0r3n5iC-Sw33t13 ,

I agree with your idea of implementing a TODO list, but not through file `README.md`, as per your commit 975d656

I think that @effgarces should rather use the **Projects** feature of github...

F0r3n5iC-Sw33t13:
Hi @colisee, 

Sorry if it wasn't the right place but I did not know where I had to put it.

F0r3n5iC-Sw33t13:
Hi!

Happy new year

@effgarces, is it okay with you if you gives me more right on this so we can make this happen?

Issue #245, opened 2023-10-16
Find a Time function (Date Range)


arikandler:
Dear people,
on the page "search-availability.php" there is a function "find a time". This works with the input "Date Range" only if start date and end date have a difference of 1 day or the end date is left blank. Could you change this or point out that this is necessary or that the class/function also works if start date and end date are the same day.

Thanks for the help!

PierreSerpe:
Hi, I’d like to find a empty ressource in my planning but somehow the functions does not work if I don t use the interval with the exact date. It is at the moment impossible for me to find any available ressource in the futur without looking at the planning day by day. How could I fix this?

Issue #107, opened 2022-04-22
Blocked space reservation


carlosnewmusic:
In the condensed week schedule view you can book even if the time is in the past, is blocked by the unavailability schedule, this should not be the case, you should not be able to book any time in the past or blocked by the unavailability schedule.
/Web/schedule.php
![imagen](https://user-images.githubusercontent.com/19379457/164702793-a8ec4686-4bb5-45c6-928b-df8817406504.png)

JohnVillalovos:
This issue has been resolved.

Issue #75, opened 2021-10-25
"Find a time" bug


effgarces:
As mentioned by @trandrew on gitter, every time is available for a given resource, even if already booked.  Happens if the date range starts and ends on the same date.

JohnVillalovos:
This seems resolved. If not please re-open.

Issue #71, opened 2021-09-28
Full deletion of old records


gcskoenig:
For GDPR compliance, it would be helpful if there would be a feature that allows to automatically delete old records from the database, meaning that they are not only set to be "invisible" but actually deleted.

Is such a feature in planning?

PR #297, opened 2024-01-17
Bug Fix and New Feature

ramartins02:
Implemented a script that can be used (as a cron job for example) do delete old data from database (issue #71), similar to other existing cron job scripts.

This features allows the deletion of old announcements, blackouts and reservations and their respective information.

In the config file it's possible to select which data should be deleted (can choose any of the three above) and how +old it has to be in years. This script can only be run through the console (like the others already implemented).

------------------------------------------------------
Fixed the "date range" field in the "Find A Time" feature (issue #245 and #75).

------------------------------------------------------
Removed button to create reservations in condensed schedule, if date is in the past, except for app admins (issue #107).

------------------------------------------------------
Also, the facebook login can sometimes give an sdk error (very rarely), a message is now shown to the user if it happens (message includes translations but should be checked by native speakers).

### DIFF
--- a/Controls/Dashboard/GroupUpcomingReservations.php
+++ b/Controls/Dashboard/GroupUpcomingReservations.php
@@ -20,7 +20,7 @@ class GroupUpcomingReservations extends DashboardItem implements IGroupUpcomingR
     public function PageLoad()
     {
         $this->Set('DefaultTitle', Resources::GetInstance()->GetString('NoTitleLabel'));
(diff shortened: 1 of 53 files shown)


treborin (commented):
https://github.com/LibreBooking/app/commit/8aa056c2a1f59f8dd149e07c1fcd81055e051030?diff=unified&w=0#commitcomment-140337789

PR #299, opened 2024-01-21
Update minimum PHP version

Yeicor:
`php composer.phar show --tree` outputs:

    ...
    gregwar/captcha v1.2.1 Captcha generator
    |--ext-gd *
    |--ext-mbstring *
    |  +--php >=7.1
    |--php >=5.3.0
    +--symfony/finder *
       +--php >=8.2

So the `develop` branch of this repo requires PHP >= 8.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ This is a community effort to keep the OpenSource [GPLv3](./LICENSE.md) LibreBoo
 
 ## Prerequisites
 
-- PHP 8.1 or greater
+- PHP 8.2 or greater
 - MySQL 5.5 or greater
 - Web server (Apache, IIS)
 


treborin:
Hi @Yeicor , i'm not sure if now is the right moment to require PHP 8.2. Let's see what @effgarces says about it.

Yeicor:
I don't know much about the PHP ecosystem, but when I installed the dependencies of the develop branch using composer and PHP 8.1, it gave me an error saying that PHP 8.2 is required and I couldn't load any page. It seems like the `gregwar/captcha` dependency only works with PHP 8.2. 

This is the only reason why I suggested this change. I guess the alternative would be to make this dependency optional, fork it or find an alternative that works with PHP 8.1.

effgarces:
I agree with @treborin, it's still a bit too early to require php 8.2. I will see what was the last version supporting php 8.1 and will probably fork that particular project. Having said that if anyone has a better alternative to the captcha library I'm open to switch to another one.

Issue #46, opened 2021-05-14
privacy - scheduler show label details without ressource permissions


QuasiMoko:
Hi all,
unfortunately I found another bug similar to #25 where @effgarces gratefully took some effort in.
I tried to find the issue in `SlotLabelFactory.php` but for me it seems the resource permissions receive no attention at all.

**Issue**
When self registered user login without any permissions, the user can not do new bookings etc. BUT get shown existing bookings with all details (Names, e-mail, participants etc.). Actually the user is only member of a default group which has no rights at all.
By resource permissions the default group has no rights (no "view only", no "full access"). Removing the default group doesn't change the issue.
In my case new registered users should get the detailed view on the resources/ bookings only after authorization of an admin.

**Environment**
- v2.8.5.3
- self-registration = on
- auto-add default user-group with no permissions (removing all group memberships without any permissions doesn't matter)
- user has no personal rights
- it does not matter if resources public (calendar abo) ist activated or not

```
/**
 * Privacy configuration
 */
$conf['settings']['privacy']['view.schedules'] = 'true';       			// if unauthenticated users can view schedules
$conf['settings']['privacy']['view.reservations'] = 'false';    			// if unauthenticated users can view reservations
$conf['settings']['privacy']['hide.user.details'] = 'false';    			// if personal user details should be displayed to non-administrators
$conf['settings']['privacy']['hide.reservation.details'] = 'false';			// if reservation details should be displayed to non-administrators. options are true, false, current, future, past
$conf['settings']['privacy']['allow.guest.reservations'] = 'false';			// if reservations can be made by users without a Booked account, if true this overrides schedule and resource visibility
```

Thanks a lot for your support

effgarces:
I believe you are correct, the resource permissions are not checked when filling the slots in the schedule view. From what I can gather, especially from our usage, the settings to show/hide that information, only allow to control if guest/registered can see them or if only the admin can see it, there's no in-between. Ideally the resource permissions would be checked so that those items would be auto-hidden from a user without permissions.
In our usage case, we set the permissions so that only the admin could see them (hide.reservation.details = true and hide.user.details = true) and we also added some of our inventory behind groups/schedules, to which people get added manually, afterwards they can then see/reserve those items, so using this technique we have avoided needing to control individual resources permissions.

[...]

PR #307, opened 2024-02-15
Schedule Resource Permissions

ramartins02:
Implemented a new feature that checks for user resource permissions when they want to see a reservation (issue #46). If the user doesn't have permissions to a resource, all reservation details will be hidden to him (including the schedule label and pop-up), except if he his invited, a guest or the owner of the reservation. If a reservation includes multiple resources then the user must have permissions to at least one of them to check the details. The feature takes into account both user and group permissions. If the user or group of the user doesn't have permissions to a resource but is a resource admin or schedule admin in charge of it, it will be able to see the details.

This feature is basically a more specific implementation of the hide.user.details / hide.reservation.details in the config file but instead of hiding all reservations for all users indiscriminately, it checks the logged in user and its respective resource permissions.

This feature comes with the all the translations for the existing languages, but they should be checked by native speakers.
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Also, refactored some of the code of my previous pull requests for better presentation and performance and fixed some minor errors.

### DIFF
--- a/Controls/Dashboard/PendingApprovalReservations.php
+++ b/Controls/Dashboard/PendingApprovalReservations.php
@@ -4,6 +4,8 @@ require_once(ROOT_DIR . 'Controls/Dashboard/DashboardItem.php');
 require_once(ROOT_DIR . 'Presenters/Dashboard/PendingApprovalReservationsPresenter.php');
 require_once(ROOT_DIR . 'Domain/Access/ReservationViewRepository.php');
 
+//This section of the dashboard doesn't show pending approval reservations as orange because all of them are pending
+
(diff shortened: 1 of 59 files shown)


treborin (changes requested):
File tpl/Dashboard/dashboard_reservation.tpl, line 5:
May be, a better way could be to check for var $OrangePending, if it is set to false, white color, but if it is set to true, yellow color. This way, anyone can activate/deactivate the use of yallow (could be the favorite color to use, who knows).

treborin (commented):
Spanish (Spain) translation is ok, file lang/es.php

JohnVillalovos:
@ramartins02  I think this PR caused the issue #343   Not sure if you know how to fix it. Thanks.

Issue #343, opened 2024-05-10
Class ScheduleWebServiceView contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (ISchedulePage::BindViewableResourceReservations)


JohnVillalovos:
When running LibreBooking and tried to hit the API endpoint I get this in the log:

```
[app] | [Fri May 10 13:08:51.927255 2024] [php:error] [pid 84] [client 10.89.0.25:59502] PHP Fatal error:  Class ScheduleWebServiceView contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (ISchedulePage::BindViewableResourceReservations) in /var/www/html/WebServices/SchedulesWebService.php on line 169
```

And when looking at the code it does appear the `ScheduleWebServiceView` does not implement the method `BindViewableResourceReservations`

JohnVillalovos:
My best guess is that this was caused by #307

PR #345, opened 2024-05-13
fix(api): API calls were broken due to abstract method

JohnVillalovos:
The class `ScheduleWebServiceView` had an abstract method `BindViewableResourceReservation` which was not implemented. This caused API access to fail with an error like:
```
[php:error] [pid 84] [client 10.89.0.25:59502] PHP Fatal error:  Class ScheduleWebServiceView contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (ISchedulePage::BindViewableResourceReservations) in /var/www/html/WebServices/SchedulesWebService.php on line 169
```

Implement the missing method as a `no-op` method.

Closes: #343

### DIFF
--- a/WebServices/SchedulesWebService.php
+++ b/WebServices/SchedulesWebService.php
@@ -524,6 +524,11 @@ class ScheduleWebServiceView implements ISchedulePage
         $this->available = false;
     }
 
+    public function BindViewableResourceReservations($resourceIds)
+    {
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
I'll be honest and admit that I don't know for sure if adding a `no-op` method is the complete solution. But I can now at least login using the API.

PR #353, opened 2024-05-22
Update config.dist.php

QuasiMoko:
set default logging level to 'none' instead of  'debug' it breaks the config-update process cause the default logging folder set does not exist in all environments


JohnVillalovos:
FYI: This PR probably needs to be rebased. When I go to the source branch it says:

`This branch is 1 commit ahead of, 276 commits behind LibreBooking/app:develop.`

effgarces:
> FYI: This PR probably needs to be rebased. When I go to the source branch it says:
> 
> `This branch is 1 commit ahead of, 276 commits behind LibreBooking/app:develop.`

Yes, I fully agree

effgarces:
I've added this change to the current release. Thank you for the contribution

PR #362, opened 2024-05-31
fix: Undefined array key 2

JohnVillalovos:
Was seeing the following errors in the log:
  PHP Warning:  Undefined array key 2 in /var/www/html/Domain/ReservationItemView.php on line 490

The issue was caused by `array_splice()`

After using `array_splice` then `$name_parts` is now only a single element array, as all other elements have been removed and used to generate `$firstnames`


stiftovich:
Shouldn't htis fix also applied to line 514 in Domain/ReservationItemView.php?
array_splice() is used there as well.

JohnVillalovos:
Probably. I haven't hit that code path yet.

Feel free to create a PR.

Issue #370, opened 2024-06-06
Schedule->Bookings page performance is very bad


JohnVillalovos:
When I go to `Schedule->Bookings` the performance is really bad.

Previously we were running version 2.7.7. And performance wasn't fast but it was acceptable.

Now I have filters set so I am showing one week and seven resources. And it takes almost 20 seconds for the "Loading reservations" status to disappear.

We have over 1,000 resources. And if I foolishly happen to load the page without filters the browser comes to a halt and is completely unresponsive.

@ramartins02 Not sure if this is related to some changes that were made to `Web/scripts/schedule.js`  I am trying to figure out the issue but my JavaScript skills are lacking. But I'm working on it and will see what I can figure out.

JohnVillalovos:
Update. So if I comment out `schedule.init();` [1] then the page loads just fine. So to me it points at the JavaScript being the issue. I was able to load a whole week schedule with over 1,000+ resources and the browser didn't lag.

[1] https://github.com/LibreBooking/app/blob/22c526a2a330091a6f18ade54697f64e0143ab05/tpl/Schedule/schedule.tpl#L452

JohnVillalovos:
As a note I checked out the `Web/scripts/scheduler.js` file from commit c90f62f3c18876482e39301e1c98532b36f771cb

And using that version things are much improved on performance for me. No idea if that will break something, but so far in very limited testing it is working.

JohnVillalovos:
Even with the old `Web/scripts/scheduler.js` performance is still not to the level version 2.7.7.

But version 2.7.7 is much much faster. Where the `develop` branch took around 120 seconds to load approximately 300 resources. 2.7.7 took around 5 seconds.

And 2.7.7 can load 1200 resources in 27 seconds. `develop` branch takes many minutes to load.

JohnVillalovos:
So the reason 2.7.7 (https://github.com/LibreBooking/app/commit/88d4673161a9e4c43a95d5796ea24a52dd70df96) is so much faster, is that the server side PHP will populate the reservations in the HTML that is sent down.

In the current version the JavaScript will populate the reservations and it is slow...

PR #371, opened 2024-06-06
Standard schedule lag when loading reservations

ramartins02:
Previous pull request (#323) may be causing the system to lag when there are to many reservations to load on the standard schedule (#370).

A new configuration is now available that lets the admin choose if he wishes to have the full label with all the text shown (true) or not (false).

With the full label it will take more time to load than the default 41px sized label. The admin should choose the preferred way based on the possible/max number of reservations the schedules have.

$conf['settings']['reservation']['show.full.label'] = 'false';

### DIFF
--- a/Pages/SchedulePage.php
+++ b/Pages/SchedulePage.php
@@ -214,6 +214,11 @@ interface ISchedulePage extends IActionPage
     public function BindViewableResourceReservations($resourceIds);
 
     /**
+     *
+     */
+    public function ShowFullLabel();
+
+    /**
      * @return LoadReservationRequest
      */
     public function GetReservationRequest();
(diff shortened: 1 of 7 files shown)


JohnVillalovos:
@ramartins02 Thanks for working on this.

I applied the patch locally but I am not noticing any improvements ☹️

I have three filters set which narrow it down to two systems, and showing one week.

I put in a `console.time()` and `console.timeLog()` at the start and end of the `renderEvents()` method.

This is the result. And yes on average `full.label = false` took longer for my 5 attempts.

 * full.label = true: 17022ms, 15584ms, 17014ms, 17127ms,  17209ms
 * full.label = false: 14841ms, 21112ms, 27911ms, 25358ms, 21285ms

And when you say "label", what does that mean? Does that mean the "name" of the resource?

Pages/SchedulePage.php
```diff
@@ -213,6 +213,11 @@ public function BindScheduleAvailability($availability, $tooEarly);
      */
     public function BindViewableResourceReservations($resourceIds);
 
+    /**
+     *
+     */
+    public function ShowFullLabel();
```
JohnVillalovos, line 219: I got this error:

```
Fatal error: Class ScheduleWebServiceView contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (ISchedulePage::ShowFullLabel) in
/var/www/html/WebServices/SchedulesWebService.php
on line 169
```

JohnVillalovos:
As a note I checked out the `Web/scripts/scheduler.js` file from commit c90f62f3c18876482e39301e1c98532b36f771cb

And using that version things are much improved on performance for me. No idea if that will break something, but so far in very limited testing it is working.

PR #373, opened 2024-06-10
feat: new algorithm for fast loading reservations

### LINKED ISSUE #370: Schedule->Bookings page performance is very bad

JohnVillalovos:
Add a new algorithm for fast loading reservations in the schedule.

Marking it as "experimental" and users will have to explicitly enable it.

Limitations:
  * Only supports the ScheduleStandard style. Otherwise it will use the default legacy algorithm.
  * Has not been setup for concurrent reservations. So behavior is undefined for concurrent reservation.

The performance gain is very significant when loading larger schedules.  On a one day schedule with 332 resources. The legacy algorithm took 23 seconds to "Load Reservations". While the new algorithm took 0.6 seconds. "Load Reservations" is the time for the renderEvents() method to run. It does not count the time for the web page to be fetched from the server.

The more resources / days in the schedule the larger the difference in timing becomes.

Closes: #370

### DIFF
--- a/Pages/SchedulePage.php
+++ b/Pages/SchedulePage.php
@@ -302,6 +302,7 @@ class SchedulePage extends ActionPage implements ISchedulePage
         $this->Set('UserIdFilter', $this->GetOwnerId());
         $this->Set('ParticipantIdFilter', $this->GetParticipantId());
         $this->Set('ShowWeekNumbers', Configuration::Instance()->GetSectionKey(ConfigSection::SCHEDULE, ConfigKeys::SCHEDULE_SHOW_WEEK_NUMBERS, new BooleanConverter()));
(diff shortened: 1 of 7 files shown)


JohnVillalovos:
@ramartins02 Not sure if you have any interest in this or not.

PR #380, opened 2024-07-01
fix: ldap_close with boolean parameter

oxconan:
- When destruct object, this function will be called, we will get issue if `this->_link` is False value
- Fix warning `userGroups` un-init value if `loadGroups` is False


JohnVillalovos (approved):
LGTM

PR #392, opened 2024-07-15
chore: update time formats in ja_jp.php

ymmt07039:
In Japan, the 24-hour notation is particularly preferred for scheduling.

reference:
https://en.wikipedia.org/wiki/Date_and_time_notation_in_Japan#Time


JohnVillalovos (approved):
LGTM

PR #396, opened 2024-07-25
Make request method configurable

hakito:
The default request method of recaptcha is _post_. This uses the method `file_get_contents` to execute the request. This [might be blocked](https://stackoverflow.com/questions/55978690/recaptcha-error-code-connection-failed-on-verification).

This PR makes the request method configurable. The default is still _post_. However I changed it to curl in the default config file.

I also fixed the logging.

### DIFF
--- a/config/config.dist.php
+++ b/config/config.dist.php
@@ -141,6 +141,7 @@ $conf['settings']['api']['allow.self.registration'] = 'false';
 $conf['settings']['recaptcha']['enabled'] = 'false';
 $conf['settings']['recaptcha']['public.key'] = '';
 $conf['settings']['recaptcha']['private.key'] = '';
+$conf['settings']['recaptcha']['request.method'] = 'curl'; // options are curl, post or socket. default: post
 /**
  * Email
  */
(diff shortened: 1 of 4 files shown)


config/config.dist.php
```diff
@@ -141,6 +141,7 @@
 $conf['settings']['recaptcha']['enabled'] = 'false';
 $conf['settings']['recaptcha']['public.key'] = '';
 $conf['settings']['recaptcha']['private.key'] = '';
+$conf['settings']['recaptcha']['request.method'] = 'curl'; // options are curl, post or socket. default: post
```
JohnVillalovos, line 144: This basically makes the default curl. Why not make it `post`?

hakito: Post uses `file_get_contents` which requires the option `allow_url_fopen` to be enabled. Because this could [cause security issues](https://stackoverflow.com/questions/24049534/is-allow-url-fopen-safe) and therefore it might not be enabled by default. On the other side I can hardly imagine a PHP installation without CURL.

hakito: I made post the default anyway so, that existing installations have the same behavior as before, when the setting is not added.

PR #398, opened 2024-07-25
Enforce redirection to Web directory

### LINKED ISSUE #397: Endless loop redirect in non-german browser
I accidently introduced a bug in 476e36fbfbc68d227c5d9c39af63f747464d9e46

When using a browser with non-german language and entering the root URL `http://librebooking.example.com/` it caused an endless loop in the line

https://github.com/LibreBooking/app/blob/229234e45dbb5e80f5e7bc7e2ec981161347d327/tpl/login.tpl#L171

hakito:
This PR always enforces a redirect to the _Web_ folder already in the .htaccess file. So I removed the index.php in the root folder which is probably bad practice anyway.

fixes #397

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -5,10 +5,8 @@ RewriteEngine On
 #RewriteCond %{HTTPS} off
 #RewriteRule .* https://%{HTTP_HOST}%{REQUEST_URI} [L,R=301]
 
-RewriteRule ^Web - [L,NC]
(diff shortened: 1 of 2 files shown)


geel97:
This is not working with NGINX. With NGINX now is needed to edit vhost to redirect.

LpSolit:
This seems to be broken with Apache too, when the app is installed in a subdirectory.

PR #399, opened 2024-07-26
Fix ldap server does not return email

oxconan:
Some ldap servers don't return user email 

When we Authenticate --> PopulateUser first time --> Email will return empty

We will try PopulateUser second time --> Email will return `uid@dc`


JohnVillalovos:
This fixed an issue we were seeing where we would not have an email and then when it went to write the new user to the database it would fail as the email address is required. With this fix an email is now generated and can then be stored in the database.

PR #408, opened 2024-09-01
Resurrect unit tests

hakito:
The unit tests seem to be broken for quite a long time. This makes it difficult to contribute to the project without breaking something essential.

So this PR is focused on fixing the phpunit tests. It covers the following changes.

- Add phpunit 11 to composer
- Fix the most obvious deprecation warnings
- Fix the tests and adapt them to phpunit 11
- Add some missing implementations and interface definitions
- Run tests in github workflow

I also adapted the script part in the composer.json. However without the .phive environment the `lint` call will fail because of the missing tools directory.

Actually I'd suggest to get rid of the .phive stuff.

### DIFF
--- a/.github/workflows/lint-and-analyse-php.yml
+++ b/.github/workflows/lint-and-analyse-php.yml
@@ -17,6 +17,39 @@ permissions:
   contents: read
 
 jobs:
+  phpunit:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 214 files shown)


JohnVillalovos:
I haven't reviewed but really like the idea of getting working unit tests. Thank you!

effgarces:
Great work, I really wanted to bring back unit tests, but it was really in my list os priorities, so thank you for your contribution

PR #415, opened 2024-09-15
Switch to utf8mb4

### LINKED ISSUE #405: UTF8MB4 incompatibility causes ERROR 1366 in MariaDB
One user tried to create a reservation with the title: ```Funny 🤡 title```

The clown face is 4 byte unicode character ```\xf0\x9f\xa4\xa1```

When trying to save the reservation it failed, but the user did not get a feedback for the reason.

In the log I saw the message:

> Error saving reservation: mysqli_sql_exception: Incorrect string value: '\xF0\x9F\xA4\xA1' for column librebooking.reservation_series.title at row 1 in

LibreBooking uses the charset `utf8` wich [normally corresponds to `utf8mb3`](https://mariadb.com/kb/en/unicode/) (also for mysql). 

I'd suggest to change all `utf8` fields and the mysqli connection to `utf8mb4`. Are there any counter-arguments?

hakito:
Fixes #405 

I changed the character set for the text fields in the database that made sense for me.

I left out fields that are

* Ids
* Auto generated
* urls
* Application internal values

I did only look at the fields and the current values of the database. I did not spend time investigating all the fields within the code base.

However I think all the fields where a user can enter something should be covered.

I also changed the default charset for all tables.


JohnVillalovos:
@hakito Should there have been any updates to the `create-*.sql` files?

hakito:
No, because I added the modifications in an upgrade file.

JohnVillalovos:
@hakito Thanks. As a note when I ran a from scratch install yesterday I got an error about `pending_reservations` table not existing. I can't find `pending_reservations` anywhere in the source code except for the upgrade file.

Any ideas on that?

PR #448, opened 2024-10-27
Fix Facebook login, smarty upgrade, Fix error on reservation page, when accessory is NULL on max or min.

lennert1986:
To ensure that NULL values in both MinQuantity and MaxQuantity do not trigger errors during validation, we can update the MaxQuantity check in the same way as the MinQuantity check.

I've added a check for is_null for both limits in the updated code.

Smarty works fine in 4.5

Package facebook/graph-sdk is abandoned, you should avoid using it. No replacement was suggested.

This is an unofficial version of Facebook's PHP SDK designed for PHP 7/8+. It is being maintained and tested against the newest PHP versions. You can use this in place of version 5.x of Facebook's deprecated facebook/graph-sdk package.


effgarces:
Thanks for the PR, can you please take a look at the failed test and see what is causing the issue with these changes.

PR #450, opened 2024-11-02
Mail translations 2nd attempt

### LINKED ISSUE #416: Limit languages for emails
When extending the templates with additional information, it is difficult or impossible for my client to update all translations.

We came up with the idea to restrict the languages to a certain set at least for the email templates. Users with a language setting that is not within the set, will fall back to the default translation.

I will come up with a PR soon to address this issue.

### LINKED ISSUE #424: Custom email templates saved for wrong language
The actual language is ignored and instead it saves it in the session language.

hakito:
* Remove *-custom.tpl from repo - should only be provided by the user
* Ignore custom translation files
* defaultPath is a fallback path - The name is misleading, as it's not the default.language setting
* Remove dead code
  * DisplayLocalized in SmartyPage was never called
  * DisplayLocalized in Page was only a copy, which is still called, but the languageCode parameter is not used.
* Fix save custom email template for selected language - fixes #424
* Fallback to default lang for missing custom templates ..but only when custom template is available for default lang.
* Add config description
* Change ReservationShareMail property name - It conflicts with the now protected email property
* Fix unit test for updating email template
* closes #416


hakito:
@effgarces The revert in  #441 removed also the changes for #416 and #424. So here is a PR only with the changes for the translations and no refactoring of the email system itself.

I also considered the [comment from lucs7 ](https://github.com/LibreBooking/app/issues/416#issuecomment-2417095139).

The tests are currently failing because of #451

hakito:
Now checks are green after DST adaption in New York. (See #451)

PR #456, opened 2024-11-14
Improved SAML instructions

cgutteridge:
Hi, I've been working through the SAML instructions and I've just got it working with Azure (yay). I found the instructions quite hard to read so I've changed the layout and made some changes and additions based on my experience following them. 

I can't see how to make a Pull Request to the wiki so I'm submitting it like this, but if you like the changes you might want to put them into the wiki rather than approve this pull request.


JohnVillalovos (approved):
Thanks for doing this!

I can't really review as I have never setup SAML. But I did read it and it seems to make sense.

The only thing I would suggest it to squash it down to one commit.

PR #467, opened 2024-12-12
New Styling plugin

JohnXLivingston:
This plugin can be used to customize the appearance. This commit includes some basic features, more could be added later.

Current features:
* plugin can add a CSS file (this file can depend on the $userSession)
* plugin can add CSS classes on the reservation items, that depends on the reservation item (you can for example read some custom attributes to change the appearance)

### DIFF
--- a/Domain/ReservationItemView.php
+++ b/Domain/ReservationItemView.php
@@ -69,6 +69,11 @@ interface IReservedItemView
     public function GetBorderColor();
 
     /**
+     * @return string[]
+     */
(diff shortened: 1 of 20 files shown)


JohnXLivingston:
For example, i'm using this new Styling Plugin to display a "lock" picture on some events.
This is based on a custom attribute that differentiate private and public events (public events will be published on the public website, using a custom SQL request).

image

JohnXLivingston:
Thanks for the merge!

Issue #472, opened 2025-01-12
[Suggestion] Channel to discuss future development


lucs7:
@effgarces 

Thanks for taking care of this project. Is there currently a discord channel or something similar where the future development of the project could be discussed?

We are planning to support the project further in the future with some man power and it would be great to have a place to exchange some ideas and questions.

effgarces:
Hi,
There's a chat over at gitter, but ever since all the changes over there I removed the link to it as the old links stopped working (https://gitter.im/bookedscheduler).
Discord is probably much better suited for this type of discussion, so I went ahead and created a space for LibreBooking (https://discord.gg/BMRYEe5v)

Those are amazing news, every bit of help is great. I really would love to spend more time with LibreBooking, but my day does not have enough hours and because of that development has stagnated a bit, if it weren't for some people in the community then it would only receive updates when something broke.

lukebarone:
The Discord link is invalid. See #480

effgarces:
I've updated the invite link, hopefully this won't expire: https://discord.gg/T8h2VkKT

Slenkis:
> I've updated the invite link, hopefully this won't expire: https://discord.gg/T8h2VkKT

link expired..

earlruby:
> > I've updated the invite link, hopefully this won't expire: https://discord.gg/T8h2VkKT
> 
> link expired.

Could someone post a new link?

Slenkis:
> > > I've updated the invite link, hopefully this won't expire: https://discord.gg/T8h2VkKT
> > 
> > 
> > link expired.
> 
> Could someone post a new link?

https://discord.gg/BNPwxaQ8qW

PR #473, opened 2025-01-12
Updated js event listeners to support DataTables pagination + x-editable

lucs7:
With the change to DataTables, event listeners in multi-page tables were broken due to rerendering of the DOM on page change.
The code changes attaches the event listeners to the table wrapper above the DataTable and specific function class. In addition x-editable is called after a page change.

Also the DataTable filter function is removed to not have two filter options.

@labmecanicatec please have a look when you have time as this should solve the discussed issue


labmecanicatec:
Hi @lucs7 , I'm currently reviewing your work and I think it's great that you were able to fix the issues that we couldn't.
The admin/manage_reservations.php, admin/manage_schedules.php and admin/manage_resources.php tables are working fine, but... we noticed that you used the develop branch when you should have used dev-bootstrap5, which contains eight commits that the main branch doesn't have. Also, if it's not too much trouble, we ask you to extend your work to the admin/manage_reservation_colors.php table.
We think it's more appropriate to make the commits that fix the datatables issues to dev-bootstrap5 and ask @effgarces  to merge them into the main branch.
There are no words to express how grateful we are for you supporting us on this project!

lucs7:
Hey @labmecanicatec . Sure thing, I will do that.
While you are already here, there is a new discord to coordinate the effort and understand what everyone is doing here in case you haven't already registered  #472

PR #482, opened 2025-02-08
Added keycloak as external auth provider

lucs7:
based on the scheme used for Google and Microsoft external auth, I added a keycloak provider that enables oauth through a keycloak instance.

Config for keycloak requires the following values

$conf['settings']['authentication']['keycloak.url'] = '';
$conf['settings']['authentication']['keycloak.realm'] = '';
$conf['settings']['authentication']['keycloak.client.id'] = '';
$conf['settings']['authentication']['keycloak.client.secret'] = '';
$conf['settings']['authentication']['keycloak.client.uri'] = '/Web/keycloak-auth.php';

Keycloak Configuration:

- Create (or use an existing) realm (e.g., “LibreBooking”) dedicated to your application.
- Client ID: Set to (for example) librebooking.
- Access Type: Set to confidential (this enables client authentication using a client secret).
- Standard Flow Enabled: Must be enabled (authorization code flow).
- Direct Access Grants/Implicit Flow: Disable these for enhanced security.
- to get title, phone number and organization mapping to user profile, the client scope should be set

### DIFF
--- a/Pages/LoginPage.php
+++ b/Pages/LoginPage.php
@@ -94,19 +94,24 @@ interface ILoginPage extends IPage, ILoginBasePage
     public function SetAnnouncements($announcements);
 
     /**
-     *  
+     *
      */
     public function SetGoogleUrl($URL);
 
     /**
-     *  
+     *
      */
     public function SetMicrosoftUrl($URL);
 
     /**
-     *  
+     *
      */
     public function SetFacebookUrl($URL);
+
+    /**
+     *
+     */
+    public function SetKeycloakUrl($URL);
 }
 
 class LoginPage extends Page implements ILoginPage
@@ -119,7 +124,7 @@ class LoginPage extends Page implements ILoginPage
 
         $this->presenter = new LoginPresenter($this); // $this pseudo variable of class object is Page object
         $resumeUrl = $this->server->GetQuerystring(QueryStringKeys::REDIRECT);
-        if($resumeUrl !== NULL) $resumeUrl = str_replace('&amp;&amp;', '&amp;', $resumeUrl);
+        if ($resumeUrl !== NULL) $resumeUrl = str_replace('&amp;&amp;', '&amp;', $resumeUrl);
         $this->Set('ResumeUrl', $resumeUrl);
         $this->Set('ShowLoginError', false);
         $this->Set('Languages', Resources::GetInstance()->AvailableLanguages);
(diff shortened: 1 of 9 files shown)


microbearlogist:
Hi @lucs7 and thanks for this nice addition.

You mention keycloack but I guess any Oauth provider works too?

lucs7:
I guess it depends how the auth url is constructed. I am not 100% sure if the request url is standardized but we could tes twith another provider. did you have something in mind?

microbearlogist:
I thought of two cases: Authentik and Nextcloud.

I'm just setting up LibreBooking in my workplace and I'd like to use Nextcloud as my identity provider.
(Long story short: we're not allowed to use an external identity provider for our nextcloud, so this would be the easiest for our users).


So I will give a try with Nextcloud first.

The Nextcloud Oauth endpoints are:
> Auth endpoint:
nextcloud.instance/index.php/apps/oauth2/authorize

> Token endpoint:
nextcloud.instance/index.php/apps/oauth2/api/v1/token

lucs7:
We should make sure not to mix up oauth and openid-connect.
I am not an expert here but as far as I recall openid-connect is another simplified layer on top of oauth to authenticate

For Nextcloud we have https://apps.nextcloud.com/apps/oidc with endpoint discovery which would need some furter changes

microbearlogist:
You're right, both providers exist:

[Oauth2 as a built-in function](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/oauth2.html) (it's a recent feature).
[OIDC as an application](https://apps.nextcloud.com/apps/oidc).

The OIDC App/plugin lacks a bit of stability and has no guarantee of being updated with Nextcloud forever, which is why I chose the built in Oauth2


Here's a 29.x Nextcloud instance showing OIDC App and built in Oauth2 providers parameters:
image

Bh4is:
Thank you very much for your work! This is exactly what I was looking for. Do you have any idea when this will be pushed?

microbearlogist:
After a quick and dirty test, it works fort Nextcloud Oauth2. I just did not manage yet to have functional scope.
I'll make a PR as soon as I have time to make something decent.

For those who wish to add more options:
[I just ](https://oauth2-client.thephpleague.com/) already offers a lot of php oauth2 clients integration.

The best thing would be to allow direct or easy integration of Oauth2 phpleague clients (but this would require a larger overhaul of librebooking's code and the working time that goes with it).


Edit:
@lucs7 The allow.keycloak.login option do not appears in the authentication configuration when I build the latest develop version.

carloscoa2:
Need to include this line in config.php file to enable "Sign in with keycloak" button option in admin interface:

$conf['settings']['authentication']['allow.keycloak.login'] = 'true';

Recommend include "Sign in with" text in a span or another html element too, for translate options (example: in oauth, you can fill the 'oauth2.name' attribute, for the button text. However, the "Sign in with" persists, what is not the proper for other languages that are not English).

PR #553, opened 2025-05-19
chore: make the resource name field wider in the schedule view

JohnVillalovos:
In the schedule, longer resource names would get cutoff previously.

### DIFF
--- a/Web/css/booked.css
+++ b/Web/css/booked.css
@@ -1157,7 +1157,7 @@ table.reservations td.resourcename a:hover {
   color: #3BA3D0;
 }
 table.reservations td.resdate {
-  width: 150px;
+  width: 500px;
   padding: 0 3px;
   background-color: #36648B;
   color: #F0F0F0;


labmecanicatec:
@JohnVillalovos,  This change won't have any effect because the Web/css/booked.css file isn't used; the actual file is Web/css/librebooking.css. In fact, only about six files are used in Web/css/. I think it's a good idea to clean up the files; it's a task I've been keeping pending after updating to Bootstrap 5.

JohnVillalovos:
@labmecanicatec Yeah, I wasn't sure about this. We are using this in our version but we have not yet moved the the Bootstrap 5 version. So I was thinking this might not be applicable.

labmecanicatec:
image
I think 500px is too much, it would be best to override the rule with a custom CSS.

JohnVillalovos:
We have names like this:

`XXX_XXX_XX_XXX_XX_XXX_XXXX_XXXX_XX___XXXXXX`

Best, IMHO, would be that it would automatically adjust the width based on how long things are.

I'm going to close this as I don't think it is applicable. We have plans to to move to the latest version but need to do some testing and make sure we don't run into any issues.

Thanks!

PR #570, opened 2025-05-21
Fix for url validation issue with public schedule/calendar

### LINKED ISSUE #563: Guest views are being redirected to non-existent paths
At the login page to LibreBooking a user can go to the upper right and select under "Schedule" one of two options.

Both options end up redirecting to invalid paths.

"View Schedule" should take the user to `Web/view-schedule.php` but instead gets redirected to `view-schedule.php` which doesn't exist.

@lucs7 has determined that https://github.com/LibreBooking/app/commit/673f2d36f60a96fda0a47775e57b3e5d1345c60f is where this issue was introduced.

lucs7:
* Updated logic for URI and param validation
* added unit test
* added URI validation to Schedule and Calendar page

**open tasks**

- [ ] should be tested to validate functionality of param validation works everywhere
- [ ] implementation should be not for each page but more central in routing to prevent repetition on every page
- [ ] RouteParamsKeys are only available for CALENDAR and SCHEDULE  but might be implemented for other pages/forms

Closes: #563

### DIFF
--- a/Pages/CalendarPage.php
+++ b/Pages/CalendarPage.php
@@ -32,6 +32,9 @@ class CalendarPage extends CommonCalendarPage implements ICommonCalendarPage
 
     public function ProcessPageLoad()
     {
+        URIScriptValidator::validateOrRedirect($_SERVER['REQUEST_URI'], '/calendar.php');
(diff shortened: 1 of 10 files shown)


JohnVillalovos (approved):
Thanks @lucs7 . LGTM.

PR #572, opened 2025-05-21
keep description and note edit from closing when losing focus

lucs7:
to make the Trumbowyg integration work as intended and allow html edits in the description and note field, the editing popover needs to remain open on link editing as this creates another popover.

### old behaviour

1. description editing button opened description editing form
2. click on insert link closed the description editing form
3. clicking outside the editing form closed editing form

### new behaviour

1. description editing button opened description editing form
2. click on insert link keeps the description editing form open and overlays the link editor
3. clicking outside the editing form does not close editing form

and a small fix for the URIScriptValidator.php for an error in the Debug logger

### DIFF
--- a/tpl/Admin/Resources/manage_resources.tpl
+++ b/tpl/Admin/Resources/manage_resources.tpl
@@ -1,2058 +1,1673 @@
 {include file='globalheader.tpl' InlineEdit=true DataTable=true Trumbowyg=true}
 
 <div id="page-manage-resources" class="admin-page">
-	<div class="clearfix border-bottom mb-3">
-		<div class="dropdown admin-header-more float-end">
-			<div class="btn-group btn-group-sm" role="group">
-				<button href="#" role="button" class="add-link add-resource btn btn-primary">
-					<span class="bi bi-plus-circle-fill icon add"></span>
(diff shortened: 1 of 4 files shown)


tpl/Admin/Resources/manage_resources.tpl
```diff
@@ -1,2058 +1,1673 @@
...
-						<div id="resourceGroups">{translate key=None}</div>
-			</form>
-		</div>
-		<div class="modal-footer">
-			{cancel_button}
-			{update_button form="resourceGroupForm" submit=true}
-			{indicator}
-		</div>
-	</div>
+    <div class="clearfix border-bottom mb-3">
```
JohnVillalovos: Any chance that this could be broken up so that the reformat happens in one commit, and the change happens in another one?

As it is right now it is very difficult to know what has changed.

lucs7: Sorry, this is due to the autoformatter. i will update the commit later without the reformatting

JohnVillalovos: Thanks!

JohnVillalovos (approved):
Thanks @lucs7

PR #590, opened 2025-06-09
fix: unclosed div in tpl\Dashboard\announcements.tpl

labmecanicatec:
- Added missing `</div>` to properly close an unclosed div in `[tpl\Dashboard\announcements.tpl]`.
- This fixes potential rendering issues caused by the malformed HTML structure.


JohnVillalovos:
LGTM

Thanks @labmecanicatec

PR #591, opened 2025-06-10
fix(layout): Adjust tooltip and inline element alignment for custom footer

labmecanicatec:
Fixed a layout issue that caused tooltips and inline elements to become misaligned when users customized the footer height.

This update ensures footer height changes do not affect UI element alignment by applying proper layout adjustments.


JohnVillalovos (approved):
LGTM

Thanks @labmecanicatec 👍

PR #597, opened 2025-06-13
style: in schedule view, make resource name wider

JohnVillalovos:
This increases the size of the column that the resource name is displayed in on the `Web/schedule.php` page. To get to that page chose `Schedule -> Bookings`.

Now resource names of 60 characters are no longer truncated.


JohnVillalovos:
I ended up installing a custom CSS file in: `Web/css/custom-style.css`

So don't need this PR.

PR #606, opened 2025-06-19
chore: remove unused LESS styles from Web/css/less

JohnVillalovos:
Deleted the Web/css/less folder containing legacy LESS styles that are no longer used in the current Bootstrap 5 setup.

This cleanup helps reduce project complexity and ensures consistency by using only CSS sources aligned with Bootstrap 5.


JohnVillalovos (commented):
LGTM

Thanks @labmecanicatec

PR #613, opened 2025-06-19
refactor(schedule.css): Remove @media (max-width: 1000px)

JohnVillalovos:
Removed the custom media query section from Web/css/schedule.css that targeted screens with max-width: 1000px.

Bootstrap 5's responsive utilities now handle these styles, making the custom rules redundant and simplifying the CSS.


JohnVillalovos (commented):
LGTM

Thanks @labmecanicatec

PR #614, opened 2025-06-19
chore: remove unused assets from ColorPicker and ColorBox

JohnVillalovos:
Removed legacy CSS, JavaScript, and image files related to ColorPicker and ColorBox, which were no longer in use following the migration to Bootstrap 5. This cleanup helps reduce project size and eliminate obsolete dependencies.


JohnVillalovos (commented):
LGTM

Thanks @labmecanicatec

PR #632, opened 2025-06-22
Update smarty5.5

labmecanicatec:
This PR updates the smarty/smarty dependency to version ^5.5 and applies a set of related code improvements to ensure compatibility with the new version and modern PHP standards.

Changes included:

- Updated smarty/smarty to ^5.5.
- Added missing use statement for the Smarty namespace.
- Replaced {html_image} tags with standard HTML  elements for displaying the logo and schedule images.
- Updated deprecated implode and join modifier syntax in .tpl templates to avoid PHP 8.2+ warnings.
- Removed unused custom Smarty plugins and methods no longer needed after cleanup.

Notes:

These changes are internal and do not affect the external behavior or features of the application. However, some testing around template rendering and plugin functionality is recommended to confirm full compatibility.

### DIFF
--- a/tpl/Admin/Attributes/attribute-list.tpl
+++ b/tpl/Admin/Attributes/attribute-list.tpl
@@ -38,7 +38,7 @@
 							{/if}</td>
 						{if $Category != CustomAttributeCategory::RESERVATION}
 							<td>{if $attribute->UniquePerEntity()}
-									{', '|implode:$attribute->EntityDescriptions()}
+									{$attribute->EntityDescriptions()|join:', '}
 								{else}
 									{translate key=All}
 								{/if}
@@ -46,7 +46,7 @@
 						{/if}
 						<td>
 							{if $attribute->HasSecondaryEntities()}
-								{', '|implode:$attribute->SecondaryEntityDescriptions()}
+								{$attribute->SecondaryEntityDescriptions()|join:', '}
 							{else}
 								{translate key=All}
 							{/if}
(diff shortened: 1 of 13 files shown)


tpl/Admin/Attributes/attribute-list.tpl
```diff
@@ -93,15 +93,15 @@
...
 		{else}
 			entityIds: [],
 			{/if}
-			entityDescriptions: ["{'","'|implode:$attribute->EntityDescriptions()}"],
+			entityDescriptions: ["{$attribute->EntityDescriptions()|join:'","'}"],
 			adminOnly: {$attribute->AdminOnly()},
 			{if $attribute->HasSecondaryEntities()}
-				secondaryEntityIds: ["{'","'|implode:$attribute->SecondaryEntityIds()}"],
-				secondaryEntityDescriptions: ["{'","'|implode:$attribute->SecondaryEntityDescriptions()}"],
+				secondaryEntityIds: ["{$attribute->SecondaryEntityIds()|joint:'","'}"],
```
JohnVillalovos: s/joint/join/

tpl/Admin/Attributes/attribute-list.tpl
```diff
@@ -93,15 +93,15 @@
...
+				secondaryEntityDescriptions: ["{$attribute->SecondaryEntityDescriptions()|joint:'","'}"],
```
JohnVillalovos: s/joint/join/

JohnVillalovos (changes requested):
Thanks @labmecanicatec  Overall looks good to me.

I left a couple of comments on what I think are issues in the PR.

tpl/Admin/Attributes/attribute-list.tpl
labmecanicatec: Thanks! I've corrected it — 'joint' is now replaced with 'join'.

tpl/Admin/Attributes/attribute-list.tpl
labmecanicatec: Thanks! I've corrected it — 'joint' is now replaced with 'join'.

JohnVillalovos (approved):
LGTM

Thanks @labmecanicatec

PR #643, opened 2025-06-26
style(config): reformat settings, combined config.*.php to config.example.php

lucs7:
Just a suggestion that still requires some more improvement to make the structure more ordered and have settings that belong to each other categorized together

NOTE: this will break the docker container as is copies config.dist.php or config.devel.php on empty config

When I got the time I will also relabel some categories and adjust the code

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -20,10 +20,11 @@ jobs:
         with:
           stale-issue-label: "stale"
           stale-pr-label: "stale"
-          exempt-assignees: 'effgarces'
+          # If an issue/PR has an assignee it won't be marked as stale
+          exempt-all-assignees: true
           operations-per-run: 500  # API calls per run, allowed 5000 per hour.
 
-          stale-issue-message: >
+          stale-issue-message: |
             This issue was marked stale because it has been open 60 days with no
             activity. Please remove the stale label or comment on this issue. Otherwise,
             it will be closed in 15 days.
@@ -44,9 +45,13 @@ jobs:
             We value your input. If you can help provide a fix, we'd be happy
             to keep this issue open and support your efforts.
 
+            This is documented in CONTRIBUTING.md
+
+            https://github.com/LibreBooking/app/blob/develop/CONTRIBUTING.md
+
           days-before-issue-stale: 60
           days-before-issue-close: 15
-          close-issue-message: >
+          close-issue-message: |
             This issue was closed because it has been marked stale for 15 days
             with no activity.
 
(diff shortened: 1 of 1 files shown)


config/config.example.php
```diff
@@ -2,281 +2,263 @@
...
+$conf['settings']['cache.templates'] = 'true'; // true recommended, caching template files helps web pages render faster
+$conf['settings']['use.local.js.libs'] = 'false'; // false recommended, delivers jQuery from Google CDN, uses less bandwidth
+$conf['settings']['inactivity.timeout'] = '30'; // minutes before the user is automatically logged out
+
+// UI configuration
+$conf['settings']['home.url'] = ''; // the url to open when the logo is clicked
+$conf['settings']['logout.url'] = ''; // the url to be directed to after logging out
+$conf['settings']['default.homepage'] = '1'; // the default homepage to use when new users register (1 = Dashboard, 2 = Schedule, 3 = My Calendar, 4 = Resource Calendar)
+$conf['settings']['css.extension.file'] = ''; // full or relative url to an additional css file to include. this can be used to override the default style
+$conf['settings']['css.theme'] = 'default'; //default, dimgray, dark_red, dark_green, french_blue, pastel_blue,
```
labmecanicatec: You could add // default, dimgray, dark_red, dark_green, french_blue, cake_blue, orange

config/config.example.php
```diff
@@ -2,281 +2,263 @@
...
-/**
- * Error logging
- */
-$conf['settings']['logging']['folder'] = '/var/log/librebooking/log'; //Absolute path to folder were the log will be written, writing permissions to the folder are required
-$conf['settings']['logging']['level'] = 'none'; //Set to none disable logs, error to only log errors or debug to log all messages to the app.log file
-$conf['settings']['logging']['sql'] = 'false'; //Set to true no enable the creation of and sql.log file
 
-// IN THE REDIRECT URI'S (OF THE AUTHENTICATION YOU ARE USING) YOU NEED TO ADD THE PATH FROM THE WEBSITE DOMAIN TO THE
-// WEB/GOOGLE-AUTH.PHP or WEB/FACEBOOK-AUTH.PHP or WEB/MICROSOFT-AUTH.PHP (depending on the services you are using)
-// EG: http://localhost/Web/facebook-auth.php
```
labmecanicatec: Do not delete these comments, it is important because the complete route is needed, not from the Web folder

config/config.example.php
lucs7: never forget the orange :)

config/config.example.php
lucs7: Done :)

lucs7:
I decided to kepp the two old config.*.php files with a deprecation notice

PR #653, opened 2025-07-03
Configuration Refactor & .env Integration

lucs7:
### config.*.php
- Refactors the config.php structure to support a modern return-based PHP array style
- Improves error reporting setup based on LB_APP_DEBUG

### configuration.php
- Adds support for environment variables via .env using phpdotenv
- Introduces a new env() helper function that mimics Laravel-style environment resolution
- Maintains backwards compatibility with the legacy $conf['settings'] format
- Updates Configurator class to output clean, pretty-formatted config files
- Creates timestamped backups (*.bck.php) of overwritten config files
- Adds a comment with timestamp to written config files
- Adds alert message to configuration page when .env file is found

### ConfigKeys.php
- updated format to fit categories
- added legacy support and warning message

### .env
- all settings are available as 'LB_' environment variables through environment on start or .env
- Configuration::Register now accepts an additional .env file (mostly important for testing as default .env should be in root folder)
- env vars are taken from environment or .env file and overwrite config.php settings
- .env implementation is mutable so later .envs can overwrite them (set environment vars e.g. docker will be overwritten by .env)

### plugins
- updated Configuration::Register
- updated *.config.dist.php files


JohnVillalovos:
@lucs7 Thanks.

One question. What do you think about having all of the environment variables prefixed with `LB_` ?  That way it is easy to know they are for LibreBooking and also less likely to have a conflict with some other program.

I'll work on finding some time over the weekend to review the whole PR.

lucs7:
Added the prefix 'LB_' similar to how the  docker imgage works

lucs7:
@JohnVillalovos 
Sorry for the large change in the codebase. I tried to to divide it but it is hard split the changes in the function calls from the change in the logic. Fortunately more than half the files are just single line changes from GetSectionKey to GetKey

tpl/Admin/Configuration/manage_configuration.tpl
JohnVillalovos, line 1: @lucs7 Can you look at this file. It seems like it got reformatted and then changed. Hard to review the changes.

JohnVillalovos (approved):
Thanks @lucs7 

I did my best to review this but it was very large. In the future we should try to have smaller PRs and smaller commits.

But I will approve this.

JohnVillalovos:
@lucs7 I rebased this and fixed the conflicts.

PR #654, opened 2025-07-04
BugFix: Wrong input date on reservation edit

madyel:
Fix Applied

The issue occurred because the default date was being set unconditionally, even on the reservation.php page, causing a one-day offset.

To resolve this, the following condition was added:

```
{if $smarty.server.SCRIPT_NAME|basename != 'reservation.php'}
    {if $DefaultDate}
        var defaultDate = {dateFormat date=$DefaultDate};
        $controlId.val(defaultDate);
    {/if}
{/if}
```

This ensures that the default date is only applied on pages other than reservation.php, preventing it from interfering with the existing reservation values.

As a result, the start and end dates now correctly match the selected event when editing a reservation.

### DIFF
--- a/tpl/Controls/DateSetup.tpl
+++ b/tpl/Controls/DateSetup.tpl
@@ -16,10 +16,11 @@
             $controlId.attr('max', maxDate);
         {/if}
 
-        {if $DefaultDate}
-            var defaultDate = {dateFormat date=$DefaultDate};
-            $controlId.val(defaultDate);
-
+        {if $smarty.server.SCRIPT_NAME|basename != 'reservation.php'}
+            {if $DefaultDate}
+                var defaultDate = {dateFormat date=$DefaultDate};
+                $controlId.val(defaultDate);
+            {/if}
         {/if}
 
         {if $AltId neq ''}


JohnVillalovos:
@madyel Need to squash down and get rid of the bad commits. Adding more commits doesn't fix it. As the comittizen check is checking all the commit messages of all the commits.

If you need help on squashing commits please let me know.

madyel:
@JohnVillalovos Thanks for the tip, really appreciated! You were right — adding more commits didn’t help. I squashed everything.

JohnVillalovos:
Thanks for the PR. Until others can reproduce the bug I am going to wait on merging in a fix. We need more information on how to reproduce. @lucs7 is working on investigating that.

madyel:
Thanks for the update! Let me know if there's anything else I can help with in the meantime.

lucs7:
@madyel thanks for the PR 🚀. 

I found the issue in the date to string conversion in js for any user timezone UTC+ and changed the date conversion  (#657). The issue might also occur outside of reservations so I guess its better to try to get the root cause here

JohnVillalovos:
@madyel Thanks for working on this! Really appreciate it.

But I went with @lucs7 PR #657 to resolve the issue.

madyel:
Got it - no worries! Glad the issue is resolved.

PR #656, opened 2025-07-04
fix: don't attempt to reserve view-only resource

### LINKED ISSUE #655: [BUG] Resources with read-only access appear bookable

Report from a user via email:

Hello, I'm sorry to bother you directly, I haven't a github account to send a post in discussions. First I want to apologize for my english.

I think there's a minor bug in the current version of LibreBooking, well, at least in how I think it should work.

In order for users to view reservation details, they must be given read access to resources even if they aren't allowed to book (classrooms). If this is done, the resource slots in resources with read permission appear as bookable on the scheduling page, however, if they try to do so, an error message appears indicating that they don't have permission upon confirmation.

I don't know if this behavior is intentional and then I'm wrong, but In my opinion, it would be more logical if that a user have only read access for a resource, the corresponding slots in the reservations grid would must appears as locked.

By making a small change to the reservation templates (i.e. tpl/Schedule/schedule-reservations-grid.tpl on lines 36 and 52) changing the use of the variable "CanAccess" to "CanBook", the behavior is as desired (at least in our case), users can see the reservation details, but the slots do not appear as reservable.

Thant you so much for reding me and your effort in LibreBooking

JohnVillalovos:
Previously a user could attempt to reserve a resource that they had "view-only" permission. The reservation would then fail.

Change the UI so that it doesn't offer them the option to reserve the resource.

Closes: #655

### DIFF
--- a/tpl/Schedule/schedule-reservations-grid.tpl
+++ b/tpl/Schedule/schedule-reservations-grid.tpl
@@ -33,7 +33,7 @@
                 <tr class="slots" data-resourceid="{$resource->GetId()}">
                     <td class="resourcename"
                         {if $resource->HasColor()}style="background-color:{$resource->GetColor()} !important" {/if}>
-                        {if $resource->CanAccess && $DailyLayout->IsDateReservable($date)}
+                        {if $resource->CanBook && $DailyLayout->IsDateReservable($date)}
                             <span resourceId="{$resourceId}"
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
With this PR I now see this in the UI

image


Conference Room 1 has been to to ViewOnly for the user.

Issue #651, opened 2025-07-02
Wrong input date on reservation edit


lucs7:
## Describe the bug**
Start and end date are one day off when editing a reservation.

## To Reproduce**
1. Create new event
2. Click reservation

## Expected behavior**
Begin and end date should correspond to event.

## Screenshots**

![Image](https://github.com/user-attachments/assets/5f150f4b-8d1b-44ca-8007-4305be6f773f)
![Image](https://github.com/user-attachments/assets/4cc76e9a-8da4-4a0d-9f2f-318edbb9d2ae)

## Specifications

  - LibreBooking version: 3.0.1

marbuka:
This could be fixed with moving the time label to the left (preferably bottom), as the hour start is represented by the left edge of the header cell, rather than the middle.

madyel:
Hi @lucs7, 

I resolved the issue by modifying the DateSetup.tpl file:

```
{if $smarty.server.SCRIPT_NAME|basename != 'reservation.php'}
    {if $DefaultDate}
        var defaultDate = {dateFormat date=$DefaultDate};
        $controlId.val(defaultDate);
    {/if}
{/if}

```
This ensures that the default date is not set when on reservation.php, thereby avoiding the one-day offset.

lucs7:
Great. Would you like to create a PR?

madyel:
Sure, the PR has already been created.

JohnVillalovos:
I am having a lot of difficulty reproducing this issue. So far I have been unable to reproduce it.

I have tried setting the server to TZ `Europe/Berlin` and making my user have the TZ `Europe/Berlin`.

Running on Firefox.

Issue #626, opened 2025-06-20
[Enhancement] Investigate use of js-beautify to format the `*.tpl` files


JohnVillalovos:
https://github.com/beautifier/js-beautify can be used to format Smarty template files `*.tpl`

https://github.com/smarty-php/smarty/discussions/1136#discussioncomment-13528574

Investigate using this to format the files and add it to the CI system.

An initial example command:
```
find tpl -type f -iname '*.tpl' -print0 \
  | xargs --null js-beautify --type html --templating smarty --indent-size 4 --replace --eol '\n' --end-with-newline
```

PR #657, opened 2025-07-06
Fix date and autogrow issues

lucs7:
1. fix(datepicker): prevent one-day offset caused by UTC conversion #651 
- bug in UTC conversion caused dates to be off by one day for UTC+n locales
- fixed by direct date parsing from PHP

2. fix(autogrow): removed unused dependency replaced by Bootstrap
- removed old autogrow dependency not in use anymore but causing console error


tpl/Reservation/create.tpl
JohnVillalovos, line 1: This is an enormous diff. Can the white-space changes be removed?

Controls/DatePickerSetupControl.php
```diff
@@ -43,11 +43,11 @@ public function PageLoad()
         $this->SetDefault('MinDate', null);
         $this->SetDefault('MaxDate', null);
 
-        if ($controlId == 'datepicker') {
-            $this->Display('Controls/DatePickerSetup.tpl');
-        } else {
-            $this->Display('Controls/DateSetup.tpl');
-        }
+        // if ($controlId == 'datepicker') {
```
JohnVillalovos: Any reason to not just delete the commented out lines? Just keep current line 49.

JohnVillalovos (changes requested):
@lucs7 Thanks.

Overall looks good to me except for the one file with all the white-space changes. Unsure if there is an actual change in there. I can't tell.

tpl/Reservation/create.tpl
JohnVillalovos: I think the editor used changed the end-of-line characters, which caused the big diff.

Before this PR:
```
$ file tpl/Reservation/create.tpl
tpl/Reservation/create.tpl: HTML document, ASCII text, with CRLF line terminators
```

This PR:
```
$ file tpl/Reservation/create.tpl
tpl/Reservation/create.tpl: HTML document, ASCII text
```

I need to work on doing a complete reformat of the code as one commit, so this stops happening... Related is #626

JohnVillalovos: I used `unix2dos` to convert it back and repushed.

JohnVillalovos (approved):
LGTM

@lucs7 Thanks again!

JohnVillalovos:
FYI: I updated the commit messages with the descriptions from the PR.

PR #661, opened 2025-07-08
fix(schedule): datepicker conflicts with url validation

lucs7:
Schedule page generates date as 2025-7-7 which conflicts with url validation as Date object as leading zeros are required.

The issue generates the following behaviour:
* user selects date
* user is redirected to schedule page

The PR adds leading zeros to date strings in the url

### DIFF
--- a/Web/scripts/schedule.js
+++ b/Web/scripts/schedule.js
@@ -1017,12 +1017,12 @@ function Schedule(opts, resourceGroups) {
             var ref = $(this).attr('ref');
             reservations.find('td[ref="' + ref + '"]').addClass('hilite');
         });
-     
+
          reservations.delegate('.clickres:not(.reserved)', 'mouseleave', function () {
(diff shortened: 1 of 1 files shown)


Web/scripts/schedule.js
```diff
@@ -1211,15 +1211,15 @@ function AddSpecificDate(dateText, inst) {
 }
 
 function dpDateChanged(dateText, inst) {
+    console.log(inst)
```
labmecanicatec: This line in console.log(inst) seems to be left behind for testing. Should we remove it?

JohnVillalovos: Good catch.

lucs7: Yes please 😂

JohnVillalovos (approved):
LGTM

Thanks @lucs7

PR #681, opened 2025-07-11
fix(API): stop storing multiple custom attributes of same type for Reources

### LINKED ISSUE #680: [BUG] Using the API to update a resource causes multiple CustomAttributes values to be added
When using the API to update a resource it will cause multiple CustomAttribute values to be added to the `custom_attribute_values` table in the database.

So if currently have the custom attribute "Test Number" set to 1 for the resource. And then use the API to change it to a value of 2.  In the `custom_attribute_values` table there will now be both entries stored. When using the web interface to edit the custom attribute it will delete the old value from the database table and then add new value.

JohnVillalovos:
When using the API to update a resource it would cause multiple custom attributes with the same ID to be stored in the
`custom_attribute_values` table in the database.

For example if currently have the custom attribute "Test Number" set to 1 for the resource. And then use the API to change it to a value of 2. In the `custom_attribute_values` table there will now be both entries stored. When using the web interface to edit the custom attribute it will delete the old value from the database table and then add new value.

Change it so that the old value will be deleted and the new value added.

Closes: #680

### DIFF
--- a/Domain/BookableResource.php
+++ b/Domain/BookableResource.php
@@ -1393,27 +1393,12 @@ class BookableResource implements IBookableResource
     private $_removedAttributeValues = [];
 
     /**
-     * @param $attributes AttributeValue[]|array
+     * @param AttributeValue[]|array $attributes
      */
     public function ChangeAttributes($attributes)
     {
-        $diff = new ArrayDiff($this->_attributeValues, $attributes);
-
-        $added = $diff->GetAddedToArray1();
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
I tested this using the API on over 1000 records. I read a resource, wrote the resource, and then verified the resource was the same as before. I also checked the SQL database and saw that the extra entries in `custom_attribute_values` were removed as they should be.

PR #682, opened 2025-07-11
style(manage_resources): don't default collapse CustomAttributes

JohnVillalovos:
Don't collapse the CustomAttributes by default in the manage resources page.

### DIFF
--- a/tpl/Admin/Resources/manage_resources.tpl
+++ b/tpl/Admin/Resources/manage_resources.tpl
@@ -578,7 +578,7 @@
 																			<i id="customAttributesIcon{$id}"
 																				class="bi bi-chevron-down"></i>
 																		</a>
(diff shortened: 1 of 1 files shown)


tpl/Admin/Resources/manage_resources.tpl
```diff
@@ -578,7 +578,7 @@
 																			<i id="customAttributesIcon{$id}"
 																				class="bi bi-chevron-down"></i>
 																		</a>
-																		<div id="customAttributes{$id}" class="collapse">
+																		<div id="customAttributes{$id}" class="collapse.show">
```
labmecanicatec: class="collapse show"

JohnVillalovos: Thanks! Done.

PR #684, opened 2025-07-13
fix: captchas not working

lucs7:
updated gregwar/captcha dependency to version 1.3
required freetype and jpeg libraries for GD extension


JohnVillalovos (commented):
LGTM

The commit message says something about freetype and jpeg, but I don't see any changes related to that.

PR #697, opened 2025-07-17
fix(reservation): pdf generation is not working in French

### LINKED ISSUE #584: French version of the reservation page do not completely load when a reminder is set.
When setting the language to French and activating the reminder function for reservations, the page do not completely load due a ' in the text of the translation in the javascript code responsible to create the pdf when an user wants to edit a reservation. It's then impossible to submit new changes. Version 2.8.6.2

The 2 problematic strings are in fr_fr.php:

```
$strings['ReminderBeforeStart'] = "avant l'heure de début";
$strings['ReminderBeforeEnd'] = "après l'heure de fin";
```

belcirelk:
When a reminder is set and the language is set to French, the reservation pdf can't be generated because the template does not escape for javascript after translation. In French the default string contains an apostrophe and this is breaking the javascript code. It also makes the update of the reservation impossible since the javascript code generate an error. Adding "|escape:'javascript'" to ReminderBeforeStart and ReminderBeforeEnd fix the problem. The code also fix the French translation.

Closes: #584

### DIFF
--- a/lang/fr_fr.php
+++ b/lang/fr_fr.php
@@ -411,7 +411,7 @@ class fr_fr extends en_gb
         $strings['hours'] = 'heures';
         $strings['days'] = 'jours';
         $strings['ReminderBeforeStart'] = 'avant l\'heure de début';
-        $strings['ReminderBeforeEnd'] = 'après l\'heure de fin';
+        $strings['ReminderBeforeEnd'] = 'avant l\'heure de fin';
         $strings['Logo'] = 'Logo';
         $strings['CssFile'] = 'Fichier CSS';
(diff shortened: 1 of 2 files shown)


belcirelk:
For information, my first try was the cleaner version {translate key=ReminderBeforeStart|escape:'javascript'} as suggested by copilot, but it had no effect.

JohnVillalovos (approved):
LGTM

Thanks @belcirelk

JohnVillalovos:
@belcirelk Feel free to join us in our Discord channel if you like https://discord.gg/4TGThPtmX8  Come by and say hi.

PR #763, opened 2025-07-24
fix(login): Language selector is not working due to httponly cookie

### LINKED ISSUE #747: The language selector is not working

## Describe the bug**

The language selector on the main page is not working. 

## To Reproduce**
The page is set to the default browser language, but it's impossible to change it using the select a the bottom of the login page. The page reload back the previous language. 

## Expected behavior**

User should be able to select the language he prefers. 
## Additional context**

Problem seems that "readCookie('language')" does not return anything. 

## Specifications

  - LibreBooking version 3.0.3

belcirelk:
The login page is set to the default browser language, but it's impossible to change the language using the select a the bottom of the login page. The page reload back the previous language.
The readCookie('language') returns null, the reason is that the cookie is set as HTTPonly and can not be read by javascript. The solution is to read the cookie via the tpl instead.

closes #747

### DIFF
--- a/tpl/login.tpl
+++ b/tpl/login.tpl
@@ -160,7 +160,7 @@
             window.location.href = url + $(this).val();
         });
 
-        var langCode = readCookie('{CookieKeys::LANGUAGE}');
+        var langCode = '{$smarty.cookies.language|escape:"javascript"}';
 
         if (!langCode) {
             langCode = (navigator.language + "").replace("-", "_").toLowerCase();


JohnVillalovos:
I just tried this and I'm not seeing a difference. I tried to use "Espanol" as the language and don't notice any difference.

This is after I selected "Espanol"  It still shows "English US" and content is English.

image

JohnVillalovos:
One thing I am noticing on my test instance is that this is the cookie set:

```
{
	"Response Cookies": {
		"language": {
			"expires": "2025-08-23T15:56:33.000Z",
			"httpOnly": true,
			"path": "/Web",
			"samesite": "Lax",
			"secure": true,
			"value": "es"
		}
	}
}
```

I am NOT using HTTPS on my test instance and here we can see that `secure: true`. Will need to figure out how to make the language cookie to be allowed to not be secure.

belcirelk:
Yes, the way it's done currently it will only work over HTTPS.

JohnVillalovos:
LGTM. I have a follow on fix for the cookie to change it to not be required to be HTTPS.

PR #766, opened 2025-07-26
chore: add text to the "More Resource Actions" drop-down

JohnVillalovos:
Before it was easy to miss the "..." and not realize there were more things that could be done on the Manage Resources page.

Change "..." to "More Resource Actions" for `en_us`


JohnVillalovos:
Example of how it looks:

image

PR #769, opened 2025-07-30
refactor: timezone handling and remove jstz library

labmecanicatec:
Replaced the manual timezone list in config/timezones.php with dynamic generation using DateTimeZone::listIdentifiers(). Removed the jstz.min.js library and updated register.tpl to use the native Intl.DateTimeFormat API for timezone detection, simplifying the code and reducing dependencies.


JohnVillalovos (approved):
LGTM

Thanks @labmecanicatec

PR #772, opened 2025-07-31
fix(chore): Resolve many html escape issues

### LINKED ISSUE #762: First Name and Last Name containing an apostrophe are not displayed correctly

## Describe the bug**

User First Name and Last Name containing apostrophe are not displayed correctly at many places
## To Reproduce**
1. Change the user first/last name for something like Jean D'Amour
2. The html code is displayed instead of the apostrophe. 
## Screenshots**

Here 2 examples
Image

Image

## Specifications

  - LibreBooking version 3.0.3

belcirelk:
The html characters are not (un)escaped properly in many TPLs. This fix the display of user names containing an apostrophe or html code. It also prevent the html display of the description and title of the reservation that
 can break the html pages.

closes #762

### DIFF
--- a/tpl/Admin/Reservations/manage_reservations.tpl
+++ b/tpl/Admin/Reservations/manage_reservations.tpl
@@ -230,7 +230,7 @@
 								data-bs-html="true">
 								<td class="id d-none">{$reservationId}</td>
 								<td class="user">
-									{fullname first=$reservation->FirstName last=$reservation->LastName ignorePrivacy=true}
+									{fullname first=$reservation->FirstName|unescape:'html' last=$reservation->LastName|unescape:'html' ignorePrivacy=true}
 								</td>
 								<td class="resource">{$reservation->ResourceName}
 									{if $reservation->ResourceStatusId == ResourceStatus::AVAILABLE}
@@ -244,8 +244,8 @@
 										{*<span class="reservationResourceStatusReason">{$StatusReasons[$reservation->ResourceStatusReasonId]->Description()}</span>*}
 									{*{/if}*}
 								</td>
-								<td class="reservationTitle">{$reservation->Title}</td>
-								<td class="description">{$reservation->Description}</td>
+								<td class="reservationTitle">{$reservation->Title|escape:'html'}</td>
+								<td class="description">{$reservation->Description|escape:'html'}</td>
 								<td class="date">
 									{formatdate date=$reservation->StartDate timezone=$Timezone key=short_reservation_date}
 								</td>
--- a/tpl/Ajax/respopup.tpl
+++ b/tpl/Ajax/respopup.tpl
@@ -56,7 +56,7 @@
 
         {capture "title"}
             {if !$hideDetails && $isResourcePermitted}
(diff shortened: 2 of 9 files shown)


JohnVillalovos:
Thanks @belcirelk  I think this needs to be rebased with the current `develop` branch. And may want to then double check `Pages/Ajax/AutoCompletePage.php` as I think it has been modified since.

tpl/Admin/Reservations/manage_reservations.tpl
```diff
@@ -244,8 +244,8 @@
 										{*<span class="reservationResourceStatusReason">{$StatusReasons[$reservation->ResourceStatusReasonId]->Description()}</span>*}
 									{*{/if}*}
 								</td>
-								<td class="reservationTitle">{$reservation->Title}</td>
-								<td class="description">{$reservation->Description}</td>
+								<td class="reservationTitle">{$reservation->Title|escape:'html'}</td>
+								<td class="description">{$reservation->Description|escape:'html'}</td>
```
JohnVillalovos, line 248: Should this be `unescape`?

tpl/Ajax/respopup.tpl
```diff
@@ -56,7 +56,7 @@
 
         {capture "title"}
             {if !$hideDetails && $isResourcePermitted}
-                <div class="title">{if $title neq ''}{$title}{else}{translate key=NoTitleLabel}{/if}</div>
+                <div class="title">{if $title neq ''}{$title|escape:'html'}{else}{translate key=NoTitleLabel}{/if}</div>
```
JohnVillalovos, line 59: Should this be `unescape`?

tpl/Dashboard/dashboard_reservation.tpl
```diff
@@ -5,9 +5,9 @@
 <div class="reservation row gx-0 {$class} border-bottom p-2 border-bottom align-items-center {if isset($orangePending)}bg-white{/if}"
     id="{$reservation->ReferenceNumber}" data-bs-custom-class="respopup-tooltip" data-bs-html="true">
     {*doesn't show pending approval reservations as orange in the Pending Approval Reservations displayer in the dashboard*}
-    <div class="col-sm-3 col-12">{$reservation->Title|default:$DefaultTitle}</div>
+    <div class="col-sm-3 col-12">{$reservation->Title|escape:'html'|default:$DefaultTitle}</div>
```
JohnVillalovos, line 8: Should this be `unescape`?

JohnVillalovos (commented):
I'm a bit confused on why sometimes `unescape` and other times `escape`.

tpl/Admin/Reservations/manage_reservations.tpl
belcirelk: The thing is in the database everything related to user is already escaped. It's not the case for resources. 

So sometime in the code the escaped strings are escaped a second time for users and not escaped at all for reservation title and description.

tpl/Ajax/respopup.tpl
belcirelk: escape to prevent the display of html in that specific case

JohnVillalovos (approved):
Thanks @belcirelk

PR #773, opened 2025-07-31
Fix for reminders on cancelled reservations

cgutteridge:
fix to AND OR logic to ensure booking-end-reminders are not sent for cancelled reservations

without the added brackets the `rs`.`status_id` <> 2 only applies to slot start reminders, not slot end.


JohnVillalovos (approved):
Thanks @cgutteridge 

LGTM (Looks Good To Me)

I did fix the commit message and rebase it to get rid of the merge commit.

PR #774, opened 2025-08-01
Fix for phpunit tests and additional error check

lucs7:
Due to the ImportICalPresenterTest tests silently failing, phpunit did fail before running all tests. Now the workflow checks for the OK message in the phpunit log.

In addition this PR fixes the 

- ImportICalPresenterTest -> update to Sabre object does not count as empty
- CheckoutPresenterTest -> new Paypal state structure
- AccountRequestValidatorTest -> EmailValidator null check 

Plugin tests have been fixed by adding provision of plugin configs based on their config.dist.php files
- PluginManagerTest
- ActiveDirectoryTest
- LdapTest

ResourceRepositoryTest has been marked as broken as it is not compatible with #681 @JohnVillalovos


JohnVillalovos (approved):
LGTM

Thanks @lucs7   I cleaned up the commit messages a little.

PR #788, opened 2025-08-08
fix: Trumbowyg fails to load when use.local.js.libs is set to true

labmecanicatec:
The Trumbowyg editor file was moved, the Resizimg plugin and required icons are now loaded. The updated JavaScript includes a template to reference the new file locations. The jQuery-resizable dependency was also added to enable image resizing functionality.


JohnVillalovos (approved):
Thanks @labmecanicatec

PR #795, opened 2025-08-14
Shib Plugin - Cannot access offset type on ShibbolethConfigKeys

amikoreh:
Error: TypeError: Cannot access offset of type array on array

Ref the 'key' for each constant to look up the configured field


lucs7 (approved):

lucs7:
can you please rename the commit according to https://www.conventionalcommits.org/en/v1.0.0/
e.g. ``fix: Shib Plugin - Cannot access offset type on ShibbolethConfigKeys``

JohnVillalovos (approved):
Thanks @amikoreh 

LGTM (Looks Good To Me).

PR #816, opened 2025-08-30
fix: ignore invalid configs in ManageConfiguration

lucs7:
This pull request makes a minor improvement to the `AddSettingFromMeta` method in `ManageConfigurationPresenter.php` by adding a check to ensure that `$meta` is set before proceeding. This helps prevent potential errors if the configuration key is not found.


JohnVillalovos (approved):
LGTM

Thanks @lucs7

PR #829, opened 2025-09-11
fix(profile): Resolve loading the profile page when multiple attributes

### LINKED ISSUE #827: Profile page crashes when more than 1 custom attributes for users

[X]: I understand that the best way to resolve this issue is to solve it and propose a Pull Request.

## Describe the bug**
The profile page crashes when more than one custom attributes is set for users.

## To Reproduce**
1. Add one custom attribute for user
2. Load the profile page
3. The page load normally with the new attribute.
4. Add a second attribute.
5. The profile page does not load anymore. 

## Specifications

  - LibreBooking version 4.0.0

belcirelk:
Fix the crash of the profile page when multiple user attributes.

closes #827

### DIFF
--- a/tpl/MyAccount/profile.tpl
+++ b/tpl/MyAccount/profile.tpl
@@ -142,18 +142,11 @@
                         </div>
                     </div>
 
-                    <div class="col-12 col-sm-6">
-                        {if $Attributes|default:array()|count > 0}
-                            {control type="AttributeControl" attribute=$Attributes[0]}
-                        {/if}
-                    </div>
-
-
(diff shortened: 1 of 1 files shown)


belcirelk:
I don't understand why I have 1 failing check. I don't see the relation with my proposed changes.

JohnVillalovos:
> I don't understand why I have 1 failing check. I don't see the relation with my proposed changes.

Most likely it is a newer version of PHPstan. I will check that out. Thanks.

JohnVillalovos (approved):
LGTM

Thanks @belcirelk

PR #859, opened 2025-10-07
fix(reservation): Resolve html rendering in announcement emails

### LINKED ISSUE #858: Announcement emails do not render html

## Describe the bug**

The emails sent when creating an announcement do not render html. 
## To Reproduce**
1. Create a announcement and choose "send by email"
2. The resulting email displays html tags

## Specifications

  - develop

belcirelk:
Added html_entity_decode|url2link in the email smarty templates.

closes #858

### DIFF
--- a/lang/da_da/AnnouncementEmail.tpl
+++ b/lang/da_da/AnnouncementEmail.tpl
@@ -1,1 +1,1 @@
-{$AnnouncementText|nl2br}
+{$AnnouncementText|html_entity_decode|url2link|nl2br}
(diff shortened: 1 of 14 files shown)


belcirelk:
It's another table were the data is saved escaped. It's not clear to me why sometimes data are saved escaped and sometimes not. Should it be the same for all db tables?

nikpael:
Hi, I saw that the subject also has problems with accents.

JohnVillalovos (approved):
LGTM

Thanks @belcirelk

PR #877, opened 2025-11-05
Fixed confusing unit test for a default configuration value

barakiva:
Unit tests fails with the following output:

 description "Invalid choice should be replaced with default"

```
1) ConfigTest::testMainConfigValidation
Invalid choice should be replaced with default
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'none'
+'error'
```

However ConfigKeys.LOGGING_LEVEL default value seems to be 'error', not 'none' (which should be None either way).

```
    public const LOGGING_LEVEL = [
        'key' => 'logging.level',
        'type' => 'string',
        'default' => 'error',
        'choices' => [
            'none' => 'None',
            'error' => 'error',
            'debug' => 'debug',
        ],
```

So unless im misreading the test case (it is admittedly confusing and i'd appreciate some context) the expected value _should_ be 'error', not 'none'. Therefore i've changed it to 'error' (the default value) and the test passed. Hope I didn't mis-understanding the purpose of the test.

### DIFF
--- a/tests/Infrastructure/Config/ConfigTest.php
+++ b/tests/Infrastructure/Config/ConfigTest.php
@@ -62,7 +62,7 @@ class ConfigTest extends TestBase
             $this->assertEquals(30, $timeout, "Invalid integer should be replaced with default");
 
             $loggingLevel = $config->GetKey(ConfigKeys::LOGGING_LEVEL);
-            $this->assertEquals('none', $loggingLevel, "Invalid choice should be replaced with default");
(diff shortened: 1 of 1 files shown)


barakiva:
I've ran git --amend to my commit message to fit coventional commits form. Please re-run CI. Next time I will commit using CZ - im still new to contributing :)

JohnVillalovos (approved):
Thanks @barakiva for the contribution.

LGTM

Issue #856, opened 2025-10-02
ActiveDirectory.config.php incorrectly parsed


abaldoni:
[X]: This is a bug and not a question.

[X]: I understand that the best way to resolve this issue is to solve it and propose a Pull Request.

## Describe the bug**
The `ActiveDirectory.config.php` format is as follows:

```
return [
    'settings' => [
            'domain.controllers' => 'list,of,dcs',
            'port' => 389,
            'username' => 'username',
...
```

This file is parsed as expected: this has been verified by

```
var_dump(Configuration::Instance()->File(ActiveDirectoryConfigKeys::CONFIG_ID))
```

The `ActiveDirectoryConfigKeys.php` however specifies that all settings should be inside a 'activedirectory' section:

```
    public const DOMAIN_CONTROLLERS = [
        'key' => 'domain.controllers',
        'type' => 'string',
        'default' => '',
        'label' => 'Domain Controllers',
        'description' => 'Comma separated list of domain controllers',
        'section' => 'activedirectory'
    ];
```

since in `ActiveDirectoryOptions.php` configuration keys are retrieved by:

```
$this->GetConfig(ActiveDirectoryConfigKeys::DOMAIN_CONTROLLERS)
```

If `ActiveDirectory.config.php` is changed as follows

```
return [
    'settings' => [
         'activedirectory' => [
            'domain.controllers' => 'list,of,dcs',
            'port' => 389,
            'username' => 'username',
...
```

it will be parsed without errors but the result will be empty:

```
object(ConfigurationFile)#174 (2) { ["_values":"ConfigurationFile":private]=> array(0) { } 
	["_configKeysClass":"ConfigurationFile":private]=> string(25) "ActiveDirectoryConfigKeys" }
```

## Specifications

  - LibreBooking version: 4.0.0

soumik-com:
Adding the existing bug reports related to this so that we can reference them at once.

- https://github.com/LibreBooking/app/issues/824

- https://github.com/LibreBooking/app/issues/854

- https://github.com/LibreBooking/app/issues/782

dijoux:
As a workaround i added $configDef['section'] = null; to ActiveDirectoryOptions.php like following :

```php
    private function GetConfig($configDef, $converter = null)
    {
        $configDef['section'] = null;
        return Configuration::Instance()->File(ActiveDirectoryConfigKeys::CONFIG_ID)->GetKey($configDef, $converter);
    }
```

vittore:
[...]

PR #884, opened 2025-11-12
Fix logging to file and LDAP authentication

christiankreidl:
Hi!

This pull request fixes:

- logging to file app.log not working due to comparison of upper- and lower-case strings
- LDAP authentication not working due to problems with new config file format  
  - ConfigKeys and config file missing prefix
  - hard-coded string instead of ConfigKey used
- ActiveDirectory authentication also fixed, but could not be tested ( #856 )
- LDAP log message not an error thus changed to debug

#

### DIFF
--- a/lib/Common/Logging/Log.php
+++ b/lib/Common/Logging/Log.php
@@ -31,7 +31,7 @@ class Log
         $this->logger = new Logger('app');
         $this->sqlLogger = new Logger('sql');
 
-        $log_level = Configuration::Instance()->GetKey(ConfigKeys::LOGGING_LEVEL);
+        $log_level = strtolower(Configuration::Instance()->GetKey(ConfigKeys::LOGGING_LEVEL));
 
         $log_folder = null;
         $log_sql = false;
@@ -103,7 +103,7 @@ class Log
      */
     public static function Error($message, $args = [])
     {
-        $log_level = Configuration::Instance()->GetKey(ConfigKeys::LOGGING_LEVEL);
+        $log_level = strtolower(Configuration::Instance()->GetKey(ConfigKeys::LOGGING_LEVEL));
         if ($log_level == 'none') {
             return;
         }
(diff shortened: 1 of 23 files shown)


christiankreidl:
fix: commit messages changed to fix linting errors

lucs7 (commented):
@christiankreidl great PR thank you for fixing this.
I would suggest a slightly different solution for the log.php (see below).

for the issue with the Plugin Configuration requiringing he plugin name in the key, I will look into a more generalized solution, so I would suggest to remove commit 58a0ba94ef7c3b49d401961281b9c1b50c7c614b for now from the PR

lib/Common/Logging/Log.php
lucs7:
```suggestion
        if ($this->log_level != 'none') {
```

lucs7:
```suggestion
            switch ($this->log_level) {
```

lucs7:
```suggestion

```

lib/Common/Logging/Log.php
```diff
@@ -103,7 +103,7 @@ public static function Debug($message, $args = [])
      */
     public static function Error($message, $args = [])
     {
-        $log_level = Configuration::Instance()->GetKey(ConfigKeys::LOGGING_LEVEL);
+        $log_level = strtolower(Configuration::Instance()->GetKey(ConfigKeys::LOGGING_LEVEL));
```
lucs7:
```suggestion

```

lucs7:
I created https://github.com/lucs7/librebooking/commit/7851a071a1aa4dc4dc44daccab5f132d23f9b029 to fix the broken PluginConfig structure, which then will correctly apply the section a

@christiankreidl Can you please test if this works with the plugins
I added the rest of the suggestions as a PR in your repo
https://github.com/christiankreidl/librebooking/pull/1

I created two docker test env to run the ActiveDiretory and Ldap plugins for login, oth passing after some testing :)

christiankreidl:
> @christiankreidl Can you please test if this works with the plugins I added the rest of the suggestions as a PR in your repo [christiankreidl#1](https://github.com/christiankreidl/librebooking/pull/1)

@lucs7  The above pull request was tested with LDAP and works. Thanks!

lucs7:
Can you merge the PR into your repo, this should update this PR too as far as I understand git ^^

lucs7:
Looks good to me 🥳

lucs7:
can be closed as #899 is taking it over

PR #888, opened 2025-11-21
Loading Libraries per default via CDN, e.g. GoogleFonts is possibly problematic with european GDPR.

Steffeng5:
To ensure "Libre" in "LibreBooking" it would be better to serve JS libs by default from local :)


lucs7:
Good idea but please follow https://www.conventionalcommits.org/

Steffeng5:
> Good idea but please follow https://www.conventionalcommits.org/

Done :)

JohnVillalovos:
Thanks @Steffeng5

Sorry this fell off my radar. I will work on this over the next few days.

JohnVillalovos (approved):
@Steffeng5 Thanks for the PR

PR #895, opened 2025-12-04
Refactor/resource detail complete UI

alejowoods:
- Get rid of iframe and schedule preview
- reorganizing layout for a better ux: back and reserve in the same div
- add image, description, notes, custom attributes and accessories divs 
- Simpler schedule and availability view.


labmecanicatec:
Although this PR is interesting in principle, it is quite incomplete. It has many untranslated text strings, it will never load the resource images nor their associated accessories, and several other issues. It stops using the tpl/Admin/Resources/view_resources.tpl file but does not remove it, and I believe it should actually be reused.

On top of that, it breaks the appearance by using Font Awesome when LibreBooking uses Bootstrap Icons, and by changing accordion-item to card. In our case, we have almost 100 resources, and this change makes it difficult to reach the bottom of the page in the panel. We do like the current appearance of the resource list and view.

alejowoods:
> Although this PR is interesting in principle, it is quite incomplete. It has many untranslated text strings, it will never load the resource images nor their associated accessories, and several other issues. It stops using the tpl/Admin/Resources/view_resources.tpl file but does not remove it, and I believe it should actually be reused.
> 
> On top of that, it breaks the appearance by using Font Awesome when LibreBooking uses Bootstrap Icons, and by changing accordion-item to card. In our case, we have almost 100 resources, and this change makes it difficult to reach the bottom of the page in the panel. We do like the current appearance of the resource list and view.

alejowoods:
Apologies for the confusion. This PR was opened in the wrong repository by mistake.

This was meant for my fork (`alejowoods/librebooking`) where we're working on internal improvements with @lucs7, not for the official LibreBooking repository.

Closing this as it's not intended for upstream. Sorry for the noise!

PR #899, opened 2025-12-16
Fix LDAP authentication

lucs7:
originally by @christiankreidl, just rebased it so we can proceeed

Hi!

This pull request fixes:

logging to file app.log not working due to comparison of upper- and lower-case strings
LDAP authentication not working due to problems with new config file format
ConfigKeys and config file missing prefix
hard-coded string instead of ConfigKey used
ActiveDirectory authentication also fixed, but could not be tested ( https://github.com/LibreBooking/app/issues/856 )
LDAP log message not an error thus changed to debug

Closes 782, 824, 856

### DIFF
--- a/.github/workflows/lint-and-analyse-php.yml
+++ b/.github/workflows/lint-and-analyse-php.yml
@@ -96,14 +96,15 @@ jobs:
               sudo chown www-data:www-data "config/$(basename ${target})"
             fi
             if ! [ -f ${target} ]; then
-              ln -s "config/$(basename ${target})" "${target}"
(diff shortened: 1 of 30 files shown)


lucs7:
Sorry but i noticed an issue in the phpunit failing silently again, hope this fixes it now

LeSteph34:
Hello,
I confirm that this fix resolves LDAP authentication problem on my server.
Thanks to @lucs7 and @christiankreidl !

JohnVillalovos (approved):
LGTM

Thanks a lot @christiankreidl and @lucs7

Issue #114, opened 2022-05-06
Error in create-data.sql


mitradranirban:
Line 7 insert into `layouts` values (1, 'America/New_York', 0);

MySQL said: 
#1136 - Column count doesn't match value count at row 1
foreign key checks should also be disabled for sql to work

Issue #94, opened 2022-02-11
Error in mysql -u booked -p bookeddb < database_schema/create-data.sql


alex-rapp:
Hi together,
first I would like to thank everybody who contributes to keep this project alive.

I did a fresh install and faced a problem at the step where the database is setup. The command:

`mysql -u booked -p bookeddb < database_schema/create-data.sql`

produced an error:
`Error 1136 at line 7: Column count doesn't match value count at row 1`

I fixed that error by modifying that lien to:
`insert into 'layouts' (timezone) values ('yourVal/yourVal');`

Is this correct?

Best wishes
Alex

treborin:
That line should be something like this:

insert into `layouts` values (1, 'America/New_York', 0);

Firs arg is the layout_id, second the time zone, third the layout_type

![image](https://user-images.githubusercontent.com/31905622/153563747-4a9f0637-2642-4d9f-a9ad-84a80104da9b.png)

treborin:
> Dear Roberto, thanks for the answer, I can confirm it looks like this for me:
> 
> SELECT * FROM layouts; +-----------+---------------+-------------+ | layout_id | timezone | layout_type | +-----------+---------------+-------------+ | 1 | Europe/Berlin | 0 | +-----------+---------------+-------------+
> 
> When I started the first command I got the above mentioned error message and when I checked then the table only had 2 columns, the last one "layout_type" was not present. Nevertheless it is working, thanks Alex

Maybe you need to check inside /database_schema)/upgrades/2.* folders. That field was added on 2.7 update

JohnVillalovos:
This issue seems stale. Closing. Feel free to reopen if desired.

Issue #72, opened 2021-10-07
Error populating data from script database_schema/create-data.sql 


ValerioParisMitritsakis:
While running the script database_schema/create-data.sql  it tries to do the following insert:

`insert into `layouts` values (1, 'America/New_York', 0);`

However per the relevant schema of the table layouts that was created from the script atabase_schema/create-schema.sql  (as shown below) there are only two columns and the script fails.

`DROP TABLE IF EXISTS `layouts`;
CREATE TABLE `layouts` (
 `layout_id` mediumint(8) unsigned NOT NULL auto_increment,
 `timezone` varchar(50) NOT NULL,
 PRIMARY KEY (`layout_id`)
) ENGINE=InnoDB DEFAULT CHARACTER SET utf8;`

the insert statement should be changed as follows in order or it to work:

`insert into `layouts` values (1, 'America/New_York');`

ValerioParisMitritsakis:
Appologies for my late reply, it was indeed a manual installation of Booked Scheduler GPLv3 v2.8.5.4

JohnVillalovos:
This issue seems stale. Closing. Feel free to reopen if desired.

PR #900, opened 2025-12-16
Fix #894: Update manual database setup documentation

### LINKED ISSUE #894: Manual Database Setup failing on new install because of missing database field
Using 4.0.0 release, and following https://librebooking.readthedocs.io/en/latest/INSTALLATION.html#manual-database-setup

`create-schema.sql` is doing
```
...
DROP TABLE IF EXISTS `layouts`;
CREATE TABLE `layouts` (
 `layout_id` mediumint(8) unsigned NOT NULL auto_increment,
 `timezone` varchar(50) NOT NULL,
 PRIMARY KEY (`layout_id`)
) ENGINE=InnoDB DEFAULT CHARACTER SET utf8;
...
```
Then `create-data.sql` is doing
```
...
insert into `layouts` values (1, 'America/New_York', 0);
...
```

So, it give this error `ERROR 1136 (21S01) at line 7: Column count doesn't match value count at row 1`, because table has 2 fields and insert is trying to fill 3 fields.

So which file is right?
- Should `create-schema.sql` needs to have field `layout_type` for that layouts table, like it is done in `2.7/schema.sql`?
- or `create-data.sql` needs to be fixed with removing 3rd value?

Thanks

lucs7:
- Document that upgrade scripts must be run between create-schema.sql and create-data.sql
- Add warning that skipping upgrades will cause create-data.sql to fail
- Clarify that the automated installer runs: schema → upgrades → data
- Update both cPanel and privileged MySQL user instructions

Closes #894, #94, #72, #114

### DIFF
--- a/docs/source/INSTALLATION.rst
+++ b/docs/source/INSTALLATION.rst
@@ -175,10 +175,27 @@ Manual Database Setup
 | The following SQL files are available:
 | - ``create-db.sql`` - Creates the database
 | - ``create-user.sql`` - Creates the database user (optional)
-| - ``create-schema.sql`` - Creates all tables and structure
+| - ``create-schema.sql`` - Creates base table structure
+| - ``database_schema/upgrades/*/schema.sql`` - Database schema upgrades
+| - ``database_schema/upgrades/*/data.sql`` - Database data upgrades
 | - ``create-data.sql`` - Inserts initial application data
 | - ``sample-data-utf8.sql`` - Sample data for testing (optional)
 |
+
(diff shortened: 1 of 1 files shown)


JohnVillalovos (approved):
Awesome! Thanks @lucs7

JohnVillalovos:
@lucs7 Just need to update the commit message subject

lucs7:
sorry, long day ^^
now it should follow cc

JohnVillalovos (approved):

PR #904, opened 2025-12-17
Fix custom attributes in reservations not shown

lucs7:
Fix autotable PDF issue that blocked ajax call. replacement of ajax call with server side handler and fallback on issue

### DIFF
--- a/Pages/Ajax/AutoCompletePage.php
+++ b/Pages/Ajax/AutoCompletePage.php
@@ -212,7 +212,7 @@ class AutocompleteUser
 
     public function __construct($userId, $firstName, $lastName, $email, $userName, $currentCreditCount = null)
     {
-        $full = new FullName(htmlspecialchars_decode($firstName, ENT_QUOTES), htmlspecialchars_decode($lastName, ENT_QUOTES));
+        $full = new FullName(htmlspecialchars_decode($firstName ?? '', ENT_QUOTES), htmlspecialchars_decode($lastName ?? '', ENT_QUOTES));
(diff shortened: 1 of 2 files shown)


tpl/Reservation/create.tpl
```diff
@@ -324,7 +324,23 @@
             </div>
 
             <div class="order-bottom border-bottom py-2">
-                <div id="custom-attributes-placeholder"></div>
+                <div id="custom-attributes-placeholder">
```
JohnVillalovos: Is this needed? Doesn't `reservation.js` fill this?

https://github.com/LibreBooking/app/blob/bea840dd2ff194ad4316b7677b1d5bc3306cc708/Web/scripts/reservation.js#L321-L331

JohnVillalovos:
@lucs7 I merged the first two commits from this PR and rebased it. Waiting for your response on the comment I left.

tpl/Reservation/create.tpl
lucs7: yes and yes. No basically it gets filled on the serverside and then the ajax call can update it e.g. if the device on the booking page is changed

lucs7:
I will look into the copilot suggestion to add the test

tpl/Reservation/create.tpl
lucs7: another idea that does not introduce two routes and we need the ajax to work for updating anyways, lets just catch errors in the pdf generation ^^

JohnVillalovos (approved):
LGTM

Thanks @lucs7

PR #909, opened 2025-12-29
Update tests to use IANA timezone identifiers

labmecanicatec:
Replaced deprecated timezone strings like 'US/Central' and 'US/Eastern' with their IANA equivalents such as 'America/Chicago' and 'America/New_York' across test files and configuration data. This improves compatibility and future-proofs the code against deprecation of old timezone names.


JohnVillalovos (approved):
Thanks @labmecanicatec 

LGTM

PR #930, opened 2026-01-30
fix: Add null check in EnsureNull method

JensWendt:
fixes an issue with the REST API for LDAP-authenticated users users

### DIFF
--- a/lib/Application/Authentication/AuthenticatedUser.php
+++ b/lib/Application/Authentication/AuthenticatedUser.php
@@ -168,6 +168,9 @@ class AuthenticatedUser
 
     private function EnsureNull($value)
     {
+        if ($value === null) {
+            return null;
+        }
         $value = trim($value);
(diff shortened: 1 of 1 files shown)


lib/Application/Authentication/AuthenticatedUser.php
JohnVillalovos, line 1: @JensWendt Thanks for the PR. The commit message will need to be fixed before this can be merged.

I suggest the commit message subject line be changed to: `fix: add null check in EnsureNull method`

JohnVillalovos (changes requested):

lib/Application/Authentication/AuthenticatedUser.php
JensWendt: done

JohnVillalovos (approved):
LGTM

Thanks @JensWendt

PR #940, opened 2026-02-02
fix: availability edit button disappears after editing schedule

labmecanicatec:
Render availability dates as Y-m-d on the server and stop client-side date formatting. JS now reads data-start-date/data-end-date via attr() and the formatDate helper was removed. refreshAvailability targets .availabilityContent (instead of .availabilityPlaceHolder) and hides the modal after update. Template changes: manage_availability.tpl uses format='Y-m-d' for start/end attributes; manage_schedules.tpl wraps the include in .availabilityContent, adjusts the .availabilityPlaceHolder layout to use flex alignment, and moves the edit link outside the refreshed content with spacing. These changes avoid locale/parse issues and make partial DOM refreshes more reliable.


JohnVillalovos (approved):
LGTM

Thanks @labmecanicatec

PR #941, opened 2026-02-02
fix: include conditionally displaying title and description

JensWendt:
with no title or description there was a PHP (I assume 8+) warning being output directly into the JavaScript code, which creates an unescaped line break that causes a JS syntax error. Now neither Title nor Description can be NULL.


JohnVillalovos (approved):
LGTM

Thanks @JensWendt 

I will squash this down to one commit.

PR #942, opened 2026-02-02
refactor(date-helper.js): centralize midnight handling in time-range validation

labmecanicatec:
Centralize midnight handling in date-helper: treat an end time of '00:00' as '24:00' for time-difference validation. Remove the legacy midnight-special-case code from admin/quota.js and delegate validation to dateHelper.ValidateTimeRangeElements instead, preserving the previous behavior where midnight represents an interval that spans into the next day.

### DIFF
--- a/Web/scripts/admin/quota.js
+++ b/Web/scripts/admin/quota.js
@@ -80,25 +80,9 @@ function QuotaManagement(opts) {
 			return true;
 		}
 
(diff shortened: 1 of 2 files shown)


JohnVillalovos (approved):
LGTM

Thanks @labmecanicatec 

GitHub CI seems to be having some issues at the moment but will merge once things pass.

PR #977, opened 2026-02-07
ci(phplint): parallelize the phplint run

JohnVillalovos:
On my system it would previously take ~46 seconds to run, now it takes ~12 seconds.


labmecanicatec:
Previously 39 seconds, now 10 seconds

PR #992, opened 2026-02-10
feat: Generate a helpful web page if missing composer

JohnVillalovos:
If `composer install` has not been setup yet, causing `vendor/autoload.php` to not be created, generate a helpful error message on the web page.

Before the server would just generate a 500 message with no details.


JohnVillalovos:
It now looks like this:

image

PR #1036, opened 2026-02-17
fix: pdf generation for on the reservation page

JohnVillalovos:
Previously the JSON template lost its opening {, and after the recent async change the PDF flow now waits on that malformed JSON, so .done() never runs and the button appears dead.

Also add some default values if `null` values are present


JohnVillalovos:
Eric on discord tested and said it fixed the issue.

PR #1044, opened 2026-02-18
fix(config) time constraint 'same_day' config not working

### LINKED ISSUE #1043: reservation.start.time.constraint = 'same_day' is not working

## Describe the bug**

The setting parameters 'current' for reservation.start.time.constraint has been replaced to same_day in PR #653. However the new setting is not working and acts as default value. The current value is not working too. 

The reason is the same_day is not taken is consideration in Domain/Values/ReservationStartTimeConstraint.php

Also the logic has not changed and the old term current is more representative of the effect of the setting (current reservations can be changed, but not when they are finished).

## Specifications

  - LibreBooking version (or SHA if using the `develop` branch):

belcirelk:
The value 'same_day' of the config reservation.start.time.constraint is reverted to its previous value 'current'. same_day was not taken into account. The 'current' value describe more the behavior than 'same_day' since the reservations can not be changed after they are finished.

Closes #1043

### DIFF
--- a/config/config.dist.php
+++ b/config/config.dist.php
@@ -308,7 +308,7 @@ return [
             # Enable a waitlist for fully booked reservations (true/false)
             'allow.wait.list' => false,
 
-            # Restrict start times (e.g., 'future', 'none', 'same_day')
+            # Restrict start times (e.g., 'future', 'none', 'current')
             'start.time.constraint' => 'future',
 
             # Require approval when an existing reservation is updated (true/false)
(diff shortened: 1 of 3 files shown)


belcirelk:
Perhaps @lucs7 should take a look since he changed 'current' for 'same_day'.

JohnVillalovos:
Looks like the commit message needs a comma ':' after 'fix(config)'.  Should be `fix(config): time constraint 'same_day' config not working`

JohnVillalovos:
@belcirelk  Thanks. I agree that `docs/source/ADVANCED-CONFIGURATION.rst` should also be updated. Besides that I think it looks good to merge.

JohnVillalovos (approved):
Thanks @belcirelk LGTM.

I'll give @lucs7 time to review. But will plan on merging it on the 18th if I don't hear any objections.

belcirelk:
Perfect. I don't know if he had a plan to change something with that.

PR #1067, opened 2026-02-21
feat(assets): load Flatpickr globally with themed month dropdown styles

JohnVillalovos:
Add Flatpickr CSS/JS includes to the global header and JS include templates (moved out of the DatePicker-only conditional) so Flatpickr is loaded in both template branches. Add CSS rules for the Flatpickr month dropdown and month option colors to match site theme. Minor whitespace/newline fixes in CSS and JS include.


Web/css/librebooking.css
```diff
@@ -657,6 +657,16 @@ input.mid-number {
     color: var(--text-color-btn) !important;
 }
 
+.flatpickr-monthDropdown-months {
+    background-color: var(--primary);
+    color: #ffffff;
```
JohnVillalovos, line 662: The color is hard-coded instead of using a 'var'. Is that intentional? That will mean custom themes won't work, most likely.

Also should `!important` be used, as in the other uses in this file?

Web/css/librebooking.css
```diff
@@ -657,6 +657,16 @@ input.mid-number {
...
+}
+
+.flatpickr-monthDropdown-months option {
+    background-color: #ffffff;
+    color: #000000;
```
JohnVillalovos, line 667: The color is hard-coded instead of using a 'var'. Is that intentional? That will mean custom themes won't work, most likely.

Also should `!important` be used, as in the other uses in this file?

labmecanicatec (commented):
Everything is OK

Web/css/librebooking.css
labmecanicatec: image
In this case, it unifies the primary background color and text color to white; in general, it maintains contrast regardless of the color of the `--primary` variable. There is generally an excess of `!important`.

Web/css/librebooking.css
labmecanicatec: image
Here, the appearance of a normal selectlist is maintained.

Web/css/librebooking.css
JohnVillalovos: I'm going to assume it is intentional and keep it this way.

Web/css/librebooking.css
JohnVillalovos: I'm going to assume then it is intentional and keep it this way.

JohnVillalovos (commented):
Thanks @labmecanicatec 

LGTM

PR #1080, opened 2026-02-23
feat(email): add SMTP AutoTLS configuration option

### LINKED ISSUE #801: SMTP tries to use TLS on port 25, needs to send unencrypted
## Describe the bug**
When trying to send mail to from LibreBooking to our local mail relay server, which uses port 25 with no encryption, it attempts to use TLS instead.

## To Reproduce**

Using these settings in config.php:

`'mailer' => 'smtp',`
`'smtp.host' => 'relayserver.hostname',`
`'smtp.port' => 25,`
`'smtp.secure' => '',`
`'smtp.auth' => false,`
`[...]`
`'smtp.debug' => true,`

Attempting to reset a password shows this SMTP Debug:

> 2025-08-21 23:16:10 SMTP INBOUND: "250-ETRN"
> 2025-08-21 23:16:10 SMTP INBOUND: "250-STARTTLS"
> 2025-08-21 23:16:10 SMTP INBOUND: "250-ENHANCEDSTATUSCODES"
> 2025-08-21 23:16:10 SMTP INBOUND: "250-8BITMIME"
> 2025-08-21 23:16:10 SMTP INBOUND: "250-DSN"
> 2025-08-21 23:16:10 SMTP INBOUND: "250 CHUNKING"
> 2025-08-21 23:16:10 SERVER -> CLIENT: 250-relayserver.hostname250-PIPELINING250-SIZE 157286400250-VRFY250-ETRN250-STARTTLS250-ENHANCEDSTATUSCODES250-8BITMIME250-DSN250 CHUNKING
> 2025-08-21 23:16:10 CLIENT -> SERVER: STARTTLS
> 2025-08-21 23:16:10 SMTP INBOUND: "220 2.0.0 Ready to start TLS"
> 2025-08-21 23:16:10 SERVER -> CLIENT: 220 2.0.0 Ready to start TLS
> 2025-08-21 23:16:10 Connection failed. Error #2: stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages:error:0A000086:SSL routines::certificate verify failed [/var/www/html/vendor/phpmailer/phpmailer/src/SMTP.php line 504]
> SMTP Error: Could not connect to SMTP host. Connection failed. stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages:error:0A000086:SSL routines::certificate verify failed
> 2025-08-21 23:16:10 CLIENT -> SERVER: QUIT
> 2025-08-21 23:16:10
> 2025-08-21 23:16:10
> 2025-08-21 23:16:10
> 2025-08-21 23:16:10
> 2025-08-21 23:16:10 Connection: closed
> SMTP Error: Could not connect to SMTP host. Connection failed. stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages:error:0A000086:SSL routines::certificate verify failed`

## Specifications

  - LibreBooking v4.0.0 installed via Docker

victor-rds:
Add control of the PHPMailer `SMTPAutoTLS` option. This allows disabling
of attempting to do STARTTLS when connecting to an email server.

Closes: #801

### DIFF
--- a/config/config.dist.php
+++ b/config/config.dist.php
@@ -138,6 +138,9 @@ return [
             # Options: tls, ssl
             'smtp.secure' => '',
 
+            # SMTP Auto TLS, if an unencrypted SMTP connection should attempt to use STARTTLS
+            'smtp.autotls' => true,
+
             # Enable SMTP authentication (true/false)
             'smtp.auth' => true,
 
--- a/docs/source/ADVANCED-CONFIGURATION.rst
+++ b/docs/source/ADVANCED-CONFIGURATION.rst
@@ -159,6 +159,7 @@ Advanced PHPMailer Settings
    'phpmailer' => [
        'sendmail.path' => '/usr/sbin/sendmail',
        'smtp.debug' => false,
+       'smtp.autotls' => true,
    ],
 
 **phpmailer.sendmail.path**
@@ -166,6 +167,10 @@ Advanced PHPMailer Settings
 
 **phpmailer.smtp.debug**
   Enable SMTP debug output (true/false).
+
+**phpmailer.smtp.autotls**
+  Set PHPMailer's SMTPAutoTLS setting (true/false). Determines if an
+  unencrypted SMTP connection should attempt to use STARTTLS.
 
 Logging Configuration
 ---------------------
--- a/lib/Config/ConfigKeys.php
+++ b/lib/Config/ConfigKeys.php
@@ -312,6 +312,15 @@ class ConfigKeys
         'section' => 'phpmailer'
     ];
 
+    public const PHPMAILER_SMTP_AUTOTLS = [
+        'key' => 'phpmailer.smtp.autotls',
+        'type' => 'boolean',
+        'default' => true,
+        'label' => 'SMTP AutoTLS',
+        'description' => 'If an unencrypted SMTP connection should attempt to use STARTTLS',
+        'section' => 'phpmailer'
+    ];
+
     public const PHPMAILER_SMTP_AUTH = [
         'key' => 'phpmailer.smtp.auth',
         'type' => 'boolean',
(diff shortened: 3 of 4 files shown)


JohnVillalovos:
Thanks.

I would think the following should also be updated:
  * `lib/Config/ConfigKeys.php`
  * `docs/source/ADVANCED-CONFIGURATION.rst`

Also this should NOT be the default setting.

victor-rds:
> Thanks.
> 
> I would think the following should also be updated:
> 
> * `lib/Config/ConfigKeys.php`
> * `docs/source/ADVANCED-CONFIGURATION.rst`
> 
> Also this should NOT be the default setting.

Agreed, it is an edge case and shouldn't be the default, that's why I prefer to use 'none' instead of an empty string, as that way it won't affect old installations.

If you prefer another keyword like 'disabled' or 'insecure', let me know. I chose 'none' based on the discussion in #801.

I will take a look on the other files and update this PR

JohnVillalovos:
> Agreed, it is an edge case and shouldn't be the default, that's why I prefer to use 'none' instead of an empty string, as that way it won't affect old installations.
> 
> If you prefer another keyword like 'disabled' or 'insecure', let me know. I chose 'none' based on the discussion in #801.

I think `disabled` is probably a good value. To me that makes more sense. Or `insecure`. Both seem more clear than `none`.

JohnVillalovos:
Just another thought. What if added a config value to set `SMTPAutoTLS` to be either `true` or `false` and default to `true`?

JohnVillalovos:
> Just another thought. What if added a config value to set `SMTPAutoTLS` to be either `true` or `false` and default to `true`?

After thinking about this some more I do think this would be the best solution.

@victor-rds What do you think?

victor-rds:
> > Just another thought. What if added a config value to set `SMTPAutoTLS` to be either `true` or `false` and default to `true`?
> 
> After thinking about this some more I do think this would be the best solution.
> 
> @victor-rds What do you think?

Yes it is a better solution, I will learn how librebooking handles missing config keys and default values, the last thing I want is to break any instance without this setting.

JohnVillalovos (approved):
Thanks @victor-rds 

LGTM

PR #1109, opened 2026-03-06
fix(registration): Resolve a server error when more than one attributes

belcirelk:
Added a loop in the tpl to display all attributes on the registration page.

### DIFF
--- a/tpl/register.tpl
+++ b/tpl/register.tpl
@@ -173,19 +173,13 @@
                             </div>
                         {/if}
 
+                        {foreach from=$Attributes item=attribute}
                         <div class="col-12 col-sm-6">
-                            {if $Attributes|default:array()|count > 0}
-                                {control type="AttributeControl" attribute=$Attributes[0]}
-                            {/if}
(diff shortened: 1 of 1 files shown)


tpl/register.tpl
```diff
@@ -173,19 +173,11 @@
...
-                        <div class="col-12 col-sm-6">
-                            {if $Attributes|default:array()|count > 0}
-                                {control type="AttributeControl" attribute=$Attributes[0]}
-                            {/if}
+                    {foreach from=$Attributes item=attribute}
+		                <div class="col-12 col-sm-6">
+            		            {control type="AttributeControl" attribute=$attribute}
                         </div>
-
-                    </div>
```
JohnVillalovos, line 182: Seems valid that this should not have been removed.

JohnVillalovos (changes requested):

JohnVillalovos:
I think the unit tests are having problems because daylight savings time is happening in a few days.

JohnVillalovos (approved):
Thanks @belcirelk 

LGTM.  I will squash it down to one commit for merge.

PR #1120, opened 2026-03-08
chore: Upgrade jQuery UI assets to 1.14.2

labmecanicatec:
Remove legacy jQuery UI 1.12.1 assets (images, minified CSS/JS, metadata) and add jQuery UI 1.14.2 assets (minified CSS, JS, images, metadata). Also update tpl/globalheader.tpl to reference the new/updated UI assets (adjusted paths/markup as needed). This updates the bundled jQuery UI to a newer patch release and cleans up the old files.


JohnVillalovos:
What do you think about adding the source map file and the unminified JS?

Also are we sure that this is compatible with jQuery 3.3.1? Should we upgrade jQuery first?

tpl/globalheader.tpl
```diff
@@ -53,14 +53,12 @@
             {vendor_css src="datatables-buttons/2.4.2/css/buttons.bootstrap5.min.css"}
         {/if}
     {else}
-        <link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/smoothness/jquery-ui.css" type="text/css" />
+        <link rel="stylesheet" href="https://code.jquery.com/ui/1.14.2/themes/smoothness/jquery-ui.css" type="text/css" />
```
JohnVillalovos, line 56: Some AI review feedback.

The CDN CSS link for jQuery UI lacks a Subresource Integrity attribute:

```
  <!-- Missing integrity= and crossorigin= -->
  <link rel="stylesheet" href="https://code.jquery.com/ui/1.14.2/themes/smoothness/jquery-ui.css" type="text/css" />
```
The CDN JS link correctly has an integrity hash. The CSS should too. This was also missing in the old 1.12.1 link, but since this commit touches this line, it's a good time to fix it.

The SRI hash can be generated with:
```
curl -s https://code.jquery.com/ui/1.14.2/themes/smoothness/jquery-ui.css | openssl dgst -sha384 -binary | openssl base64 -A
```

PR #1122, opened 2026-03-08
chore: Upgrade jQuery UI assets to 1.14.2

labmecanicatec:
Remove legacy jQuery UI 1.12.1 assets (images, minified CSS/JS, metadata) and add jQuery UI 1.14.2 assets (minified CSS, JS, images, metadata). Also update tpl/globalheader.tpl to reference the new/updated UI assets (adjusted paths/markup as needed). This updates the bundled jQuery UI to a newer patch release and cleans up the old files.


JohnVillalovos (approved):
Thanks @labmecanicatec 

LGTM.

If you ever want to do a follow up to add the unminified JS and map files, feel free.

PR #1126, opened 2026-03-08
fix(time): make Time parsing/formatting DST-safe for time-only values

JohnVillalovos:
`Time::Parse()` and internal `Time` formatting/comparison logic were implicitly tied to the current calendar date. On DST transition days, nonexistent local times like `02:00` could normalize to `03:00`, causing schedule period boundaries and related tests to shift by one hour.

This change makes `Time` deterministic as a time-of-day value object:

- Parse time strings against a fixed anchor date (`2000-01-01`) with explicit supported formats
- Keep a fallback to legacy parsing for unexpected inputs
- Use a fixed anchor date internally for `Time::Format()` / comparison support, instead of deriving date parts from "today"

### DIFF
--- a/lib/Common/Time.php
+++ b/lib/Common/Time.php
@@ -2,6 +2,15 @@
 
 class Time
 {
+    // A date that is not a Daylight Savings Time transition day
+    private const ANCHOR_YEAR = 2000;
+    private const ANCHOR_MONTH = 1;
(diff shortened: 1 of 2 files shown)


labmecanicatec (commented):
I don't see any problem, the test runs without errors.

PR #1128, opened 2026-03-10
fix(view-reservation): Correctly display the duration of the reservation

labmecanicatec:
image

Change hidden entry IDs for start/end dates from formattedBeginDate/formattedEndDate to BeginDate/EndDate (values ​​still use {formatdate... key=system}). The span for the display ReservationLength tag is uncommented and corrected. No other logic changed.


JohnVillalovos (approved):
Thanks @labmecanicatec

PR #1161, opened 2026-03-17
fix(translation): Fix date formats for italian

### LINKED ISSUE #1158: Date problem in Italian language in "find a time" view
[]: This is a bug and not a question.

[]: I understand that the best way to resolve this issue is to solve it and propose a Pull Request.

## Describe the bug**

In "find a time" view, under Italian Locale, instead of the month name it appears $1$2$3

## Screenshots**

Image

## Specifications

  - LibreBooking version (or SHA if using the `develop` branch):

belcirelk:
The en_gb is now used as the default instead of en_us to correctly set the date formats. _loadDates() is removed from the italian translation since it's a duplicate of the one set in en_gb.

Closes: #1158

### DIFF
--- a/lang/it_it.php
+++ b/lang/it_it.php
@@ -1,47 +1,13 @@
 <?php
 
 require_once('Language.php');
-require_once('en_us.php');
+require_once('en_gb.php');
 
-class it_it extends en_us
+class it_it extends en_gb
 {
     public function __construct()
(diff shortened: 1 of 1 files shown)


JohnVillalovos (approved):
Thanks @belcirelk 

LGTM.

PR #1196, opened 2026-03-21
fix(email): populate resource custom attribute values when loading re…

### LINKED ISSUE #1194: Resource custom attributes values are not added in email when a copy is sent via the reservation page.

## Describe the bug**

When you create a reservation with custom attributes, the values appear in the email the user receive. 
When you open the same reservation on librebooking and use the More -> Email button to send a copy, the resource custom attributes values are missing (the label is showing only). 

## Screenshots**

Image

Image

## Specifications

  - LibreBooking version ( `develop` branch):

belcirelk:
…servation from DB

When loading an existing reservation series via ReservationRepository (e.g. the Email button flow), BookableResource objects were created with empty attributes. This caused all resource custom attribute values to appear blank in reservation emails, even though the attribute labels were visible.

Fix: collect resources in PopulateResources() and run GetAttributeValuesCommand for each one, mirroring the logic already used in ResourceRepository::LoadById(). This ensures attribute values are populated.

Closes: #1194

### DIFF
--- a/Domain/Access/ReservationRepository.php
+++ b/Domain/Access/ReservationRepository.php
@@ -351,7 +351,7 @@ class ReservationRepository implements IReservationRepository
             $resources[] = $resource;
         }
         $reader->Free();
-        
+
         // get all reservation resources attributes
         foreach ($resources as $resource) {
             $attributeCommand = new GetAttributeValuesCommand(
--- a/Domain/Access/ReservationRepository.php
+++ b/Domain/Access/ReservationRepository.php
@@ -340,7 +340,7 @@ class ReservationRepository implements IReservationRepository
         // get all reservation resources
         $getResourcesCommand = new GetReservationResourcesCommand($series->SeriesId());
         $reader = ServiceLocator::GetDatabase()->Query($getResourcesCommand);
-        $resources = [];
+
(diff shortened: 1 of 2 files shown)


Domain/Access/ReservationRepository.php
```diff
@@ -340,15 +340,33 @@ private function PopulateResources(ExistingReservationSeries $series)
...
             if ($row[ColumnNames::RESOURCE_LEVEL_ID] == ResourceLevel::Primary) {
                 $series->WithPrimaryResource($resource);
             } else {
                 $series->WithResource($resource);
             }
+            $resources[] = $resource;
         }
         $reader->Free();
+
+        // get all reservation resources attributes
```
JohnVillalovos: Could this all be put inside the while loop on line 344? Then don't need to have a new `$resources` array.

belcirelk: most likely yes. will update.

belcirelk: Done. It was just easier to visualize for me the other way.

JohnVillalovos (approved):
Thanks @belcirelk 

LGTM

PR #1214, opened 2026-03-23
feat(i18n): add enabled.languages config to restrict language selector

### LINKED ISSUE #1213: enhancement: allow restricting available languages via config.php
## Problem

The language selector added in #1210 displays all languages defined in `lang/AvailableLanguages.php`. To restrict which languages appear, administrators must edit that PHP source file directly. This is problematic because:

1. **Upgrades overwrite customizations** — `AvailableLanguages.php` is a tracked source file, so updates to LibreBooking will clobber any changes the administrator made.
2. **Mixes configuration with code** — which languages to offer is a deployment decision, not an application concern. Every other deployment setting lives in `config/config.php`.
3. **Error-prone** — if an admin removes a language from `AvailableLanguages.php` but forgets to update `default.language` in `config.php`, the default language becomes invalid.

## Proposed Solution

Add a new config key, e.g. `available.languages`, to `config/config.php`:

```php
// All languages available (default — backwards compatible)
'available.languages' => '',

// Only these languages shown in the selector
'available.languages' => 'en_us,es,fr_fr',
```

In `Resources::LoadAvailableLanguages()`, after loading the full list from `AvailableLanguages.php`, filter it down to the configured subset. If the key is empty or absent, all languages remain available.

### Validation

- On load, if `default.language` is not in the `available.languages` list, log a warning and fall back to the first available language.
- The language selector dropdown and the `ChangeLanguagePresenter` already validate against `AvailableLanguages` on the `Resources` object, so they would automatically respect the filtered list with no changes needed.

## Benefits

- Survives upgrades — configuration lives in `config.php` which is `.gitignore`d
- Consistent with how every other setting works
- `AvailableLanguages.php` remains the master registry of *supported* languages (developer concern), while `config.php` controls which are *enabled* (admin concern)
- Simple to implement — one small change in `Resources::LoadAvailableLanguages()` plus a new `ConfigKeys` constant

## Related

- #1210 — added the language selector to the navbar

JohnVillalovos:
Add a new config key `enabled.languages` that allows administrators to restrict which languages appear in the language selector dropdown via config.php, without editing source files.

When set to a comma-separated list of language codes (e.g. "en_us,fr_fr,de_de"), only those languages are shown. When empty (the default), all supported languages are shown.

Error handling:
- Unknown language codes are logged as errors and ignored
- If all codes are invalid, falls back to all languages
- If default.language is not in the enabled list, falls back to en_us and logs an error

Also fixes a pre-existing bug where GetLanguageCode() did not validate the cookie or config default language against the supported language list, which could crash the app if either was invalid.

Closes: #1213

### DIFF
--- a/lib/Common/Resources.php
+++ b/lib/Common/Resources.php
@@ -130,6 +130,8 @@ class Resources implements IResourceLocalization
      */
     public function IsLanguageSupported($languageCode)
     {
+        $languageCode = strtolower(trim($languageCode ?? ''));
+
         return !empty($languageCode) &&
             (array_key_exists($languageCode, $this->AvailableLanguages) &&
                 file_exists($this->LanguageDirectory . $this->AvailableLanguages[$languageCode]->LanguageFile));
@@ -255,7 +257,7 @@ class Resources implements IResourceLocalization
     private function GetLanguageCode()
     {
         $cookie = ServiceLocator::GetServer()->GetCookie(CookieKeys::LANGUAGE);
-        if ($cookie != null) {
+        if ($cookie != null && $this->IsLanguageSupported($cookie)) {
             return $cookie;
         }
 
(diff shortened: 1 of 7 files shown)


labmecanicatec (approved):

lib/Common/Resources.php
```diff
@@ -264,12 +266,66 @@ private function GetLanguageCode()
...
+            return;
+        }
+
+        // Filter to only the languages specified in the config
+        $enabledCodes = array_map('trim', explode(',', $enabledConfig));
+        $filtered = [];
+
+        foreach ($enabledCodes as $code) {
+            $code = strtolower($code);
+            if ($code === '') {
```
belcirelk, line 295: You could use array_filter to remove them on line 290.

lib/Common/Resources.php
```diff
@@ -264,12 +266,66 @@ private function GetLanguageCode()
...
+        $enabledConfig = Configuration::Instance()->GetKey(ConfigKeys::ENABLED_LANGUAGES);
+
+        // If no enabled.languages configured, all supported languages are available
+        if (empty($enabledConfig)) {
+            $this->AvailableLanguages = $allLanguages;
+            return;
+        }
+
+        // Filter to only the languages specified in the config
+        $enabledCodes = array_map('trim', explode(',', $enabledConfig));
```
belcirelk, line 290: Should we check for doubles? Also maybe array_filter to remove empty of trailing coma.

lib/Common/Resources.php
```diff
@@ -264,12 +266,66 @@ private function GetLanguageCode()
...
+            $defaultFallback = ConfigKeys::DEFAULT_LANGUAGE['default'];
+            Log::Error(
+                'default.language "%s" is not in enabled.languages list, falling back to "%s"',
+                $defaultLanguage,
+                $defaultFallback
+            );
+            $filtered[$defaultFallback] = $allLanguages[$defaultFallback];
+        }
+
+        // Preserve the display order from AvailableLanguages.php
```
belcirelk: It should not be the order that the person enters in the config?

lib/Common/Resources.php
JohnVillalovos: Could do that, but this is more readable for me.

belcirelk (approved):
The code looks good and useful. Few comments comments to consider or not.

lib/Common/Resources.php
JohnVillalovos: Not necessary. If someone writes 'en_us,en_us,fr_fr', the second 'en_us' just overwrites the same key in `$filtered` with the same value — array_key_exists returns true and `$filtered[$code] = $allLanguages[$code]` is assigned again identically.

lib/Common/Resources.php
JohnVillalovos: Good question. I was thinking to maintain the original order. But actually being able to customize the order seems like a good side-effect. I'll change it.

PR #1223, opened 2026-03-25
fix(config): preserve existing config values when merging new keys fr…

### LINKED ISSUE #1222: config.php is reverted to default when a new setting is added (manage_configuration.php)

## Describe the bug

Every visit to `manage_configuration.php` overwrites the entire `config.php` with default values from `config.dist.php`, reverting all custom configuration settings.

The bug only triggers when `config.dist.php` has a key that doesn't exist in `config.php` — for example after a LibreBooking update that introduces a new configuration key. Once triggered, the merge overwrites **all** existing values, not just the missing key.

## To Reproduce

1. Configure LibreBooking with custom settings (admin email, timezone, app title, etc.)
2. Update LibreBooking to a version that adds a new key to `config.dist.php`
3. Log in as an administrator
4. Go to Admin → Configuration (`manage_configuration.php`)
5. Check `config.php` — all custom values have been replaced with defaults from `config.dist.php`

Note: if `config.php` already has all the same keys as `config.dist.php`, `IsConfigOutOfDate()` returns false, no merge happens, and the bug does not occur.

## Expected behavior

Visiting the configuration page should only add **new keys** that are missing from `config.php` (introduced in newer versions of `config.dist.php`). Existing custom values should never be overwritten.

## Additional context

The bug is in `lib/Config/Configurator.php` in the `MergeValue()` method. When a config key has metadata and no special handling rules (no environment variable, not a private field), the method returns `$newValue` (the dist default) instead of `$currentValue` (the user's existing value):

```php
private function MergeValue($key, $currentValue, $newValue)
{
    $meta = $this->GetKeyMetadata($key);
    if (!$meta) {
        return $newValue; // should be $currentValue
    }
    // ...
    return $newValue; // should be $currentValue
}
```

The fix is to return `$currentValue` as the default in all cases, since the purpose of `BringConfigFileUpToDate()` is only to add missing keys, not overwrite existing values.

## Specifications

- LibreBooking version (or SHA if using the `develop` branch): `develop`

belcirelk:
…om config.dist.php

BringConfigFileUpToDate() was calling BuildConfig() which used MergeValue() to return $newValue as the default for all keys, causing every existing config value to be overwritten with dist defaults whenever a new key was added to config.dist.php.

Added a $preserveExisting parameter (default false) to BuildConfig(), MergeArrays(), and MergeValue(). When true, MergeValue() always returns $currentValue, ensuring existing values are preserved during automatic config file updates. The form save path is unaffected since it uses the default value of false.

GetMerged() now passes $preserveExisting = true to BuildConfig() so that BringConfigFileUpToDate() only adds missing keys without touching existing values.

Closes: #1222

### DIFF
--- a/lib/Config/Configurator.php
+++ b/lib/Config/Configurator.php
@@ -17,10 +17,17 @@ interface IConfigurationSettings
      *
      * @param array $currentSettings
      * @param array $newSettings
-     * @param bool $keepMissingKeys  If true, retains keys that exist only in current config.
+     * @param bool $keepMissingKeys   If true, retains keys that exist only in current config.
+     * @param bool $preserveExisting  If true, keeps existing values in $currentSettings for keys that are also present in $newSettings,
+     *                                adding only new keys from $newSettings without overwriting user-defined values.
      * @return array
      */
-    public function BuildConfig($currentSettings, $newSettings, $keepMissingKeys = false);
+    public function BuildConfig(
+        array $currentSettings,
+        array $newSettings,
+        bool $keepMissingKeys = false,
+        bool $preserveExisting = false
+    );
 
     /**
      * Writes the given configuration array to a PHP config file.
@@ -149,7 +156,12 @@ class Configurator implements IConfigurationSettings
     {
         $currentSettings = $this->GetSettings($configPhp);
         $newSettings = $this->GetSettings($distPhp);
-        $settings = $this->BuildConfig($currentSettings, $newSettings, true);
+        $settings = $this->BuildConfig(
+            currentSettings: $currentSettings,
+            newSettings: $newSettings,
+            keepMissingKeys: true,
+            preserveExisting: true
+        );
         return [Configuration::SETTINGS => $settings];
     }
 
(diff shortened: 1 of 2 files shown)


lib/Config/Configurator.php
```diff
@@ -149,7 +149,7 @@ private function GetMerged($configPhp, $distPhp)
     {
         $currentSettings = $this->GetSettings($configPhp);
         $newSettings = $this->GetSettings($distPhp);
-        $settings = $this->BuildConfig($currentSettings, $newSettings, true);
+        $settings = $this->BuildConfig($currentSettings, $newSettings, true, true);
```
JohnVillalovos: Can this use named arguments?

```
$settings = $this->BuildConfig(currentSettings: $currentSettings, newSettings: $newSettings, keepMissingKeys: true, preserveExisting: true);
```

lib/Config/Configurator.php
```diff
@@ -20,7 +20,7 @@ public function GetSettings($file);
      * @param bool $keepMissingKeys  If true, retains keys that exist only in current config.
      * @return array
      */
-    public function BuildConfig($currentSettings, $newSettings, $keepMissingKeys = false);
+    public function BuildConfig($currentSettings, $newSettings, $keepMissingKeys = false, $preserveExisting = false);
```
JohnVillalovos: Please add a PHPDoc or a type-hint to the function. Either is fine.

Here is a type-hint version.
```
public function BuildConfig(
    array $currentSettings,
    array $newSettings,
    bool $keepMissingKeys = false,
    bool $preserveExisting = false
);
```

lib/Config/Configurator.php
```diff
@@ -197,15 +197,15 @@ public function GetSettings($file)
      * @param bool $keepMissingKeys Whether to preserve keys missing in new settings.
      * @return array Final merged config with preserved order and special handling.
      */
-    public function BuildConfig($currentSettings, $newSettings, $keepMissingKeys = false)
+    public function BuildConfig($currentSettings, $newSettings, $keepMissingKeys = false, $preserveExisting = false)
```
JohnVillalovos: Update PHPDoc or use type-hints

lib/Config/Configurator.php
```diff
@@ -197,15 +197,15 @@ public function GetSettings($file)
...
     {
-        return $this->MergeArrays($currentSettings, $newSettings, $keepMissingKeys);
+        return $this->MergeArrays($currentSettings, $newSettings, $keepMissingKeys, '', $preserveExisting);
     }
 
     /**
      * Recursively merges two arrays, preserving order and applying special logic.
      */
-    private function MergeArrays($current, $new, $keepMissing = false, $keyPrefix = '')
+    private function MergeArrays($current, $new, $keepMissing = false, $keyPrefix = '', $preserveExisting = false)
```
JohnVillalovos: Update PHPDoc or use type-hints

lib/Config/Configurator.php
```diff
@@ -242,8 +242,11 @@ private function MergeArrays($current, $new, $keepMissing = false, $keyPrefix =
     /**
      * Chooses between current and new value based on environment variables and private field rules.
      */
-    private function MergeValue($key, $currentValue, $newValue)
+    private function MergeValue($key, $currentValue, $newValue, $preserveExisting = false)
```
JohnVillalovos: Update PHPDoc or use type-hints

lib/Config/Configurator.php
```diff
@@ -197,15 +197,15 @@ public function GetSettings($file)
      * @param bool $keepMissingKeys Whether to preserve keys missing in new settings.
      * @return array Final merged config with preserved order and special handling.
      */
-    public function BuildConfig($currentSettings, $newSettings, $keepMissingKeys = false)
+    public function BuildConfig($currentSettings, $newSettings, $keepMissingKeys = false, $preserveExisting = false)
     {
-        return $this->MergeArrays($currentSettings, $newSettings, $keepMissingKeys);
+        return $this->MergeArrays($currentSettings, $newSettings, $keepMissingKeys, '', $preserveExisting);
```
JohnVillalovos: Named arguments would be nice

lib/Config/Configurator.php
```diff
@@ -218,10 +218,10 @@ private function MergeArrays($current, $new, $keepMissing = false, $keyPrefix =
 
                 if (is_array($currentValue) && is_array($newValue)) {
                     // Recursively merge nested arrays
-                    $result[$key] = $this->MergeArrays($currentValue, $newValue, true, $fullKey);
+                    $result[$key] = $this->MergeArrays($currentValue, $newValue, true, $fullKey, $preserveExisting);
```
JohnVillalovos: Named arguments would be nice

lib/Config/Configurator.php
```diff
@@ -218,10 +218,10 @@ private function MergeArrays($current, $new, $keepMissing = false, $keyPrefix =
...
                 } else {
                     // Decide whether to use current or new value
-                    $result[$key] = $this->MergeValue($fullKey, $currentValue, $newValue);
+                    $result[$key] = $this->MergeValue($fullKey, $currentValue, $newValue, $preserveExisting);
```
JohnVillalovos: Named arguments would be nice

JohnVillalovos (commented):
Thanks @belcirelk 

Overall looks good. Just some minor style cleanup suggestions.

lib/Config/Configurator.php
```diff
@@ -195,33 +207,55 @@ public function GetSettings($file)
...
      */
-    public function BuildConfig($currentSettings, $newSettings, $keepMissingKeys = false)
-    {
-        return $this->MergeArrays($currentSettings, $newSettings, $keepMissingKeys);
+    public function BuildConfig(
+        array $currentSettings,
+        array $newSettings,
+        bool $keepMissingKeys = false,
+        bool $preserveExisting = false
+    ) {
```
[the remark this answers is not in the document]
JohnVillalovos: The { is already on the same line as the closing ) — ) { — which is actually the PSR-12 style for multi-line method signatures. PSR-12 §4.4 says: when the parameter list is split across multiple lines, the closing parenthesis and opening brace go together on their own line as ) {.

The "opening brace on the next line" rule applies to single-line method declarations. For multi-line declarations, ) { on the last line is correct.

Issue #1225, opened 2026-03-25
start.time.constraint = 'future' does not prevent creating reservations in the past (v4.2.0)


Anthony-cloud-1:
When ` 'start.time.constraint' => 'future' ` is set in `config/config.php`, LibreBooking should prevent users from creating or editing any reservation whose **start time** is in the past.

However, I was able to successfully create a reservation for a time slot that had already passed (e.g. today 25 March 2026 from 07:30 to 08:00, created after 08:00).

**Expected behavior**  
The system should display a validation error such as “The start time cannot be in the past” and reject the reservation.

**Actual behavior**  
The reservation is created successfully with no error or warning.

**To Reproduce**

1. Set configuration' `config/config.php`:
   ```php
   'settings' => [
       'reservation' => [
           'start.time.constraint' => 'future',
       ],
   ],
   ```
2. Clear the template cache:
   ```bash
   rm -rf /path/to/librebooking/tpl_c/*
   ```
3. Log in (tested as admin and as regular user).
4. Try to create a reservation whose start time is clearly in the past.
5. The booking is accepted.

**Screenshots**  

<img width="1866" height="651" alt="Image" src="https://github.com/user-attachments/assets/675812ab-aa8f-4e68-b674-20bd7d92fa41" />
<img width="912" height="422" alt="Image" src="https://github.com/user-attachments/assets/11fc333c-4f00-4c9c-8ef4-6053755a0449" />
<img width="1876" height="424" alt="Image" src="https://github.com/user-attachments/assets/81a45793-4534-4a1d-adbf-73a8a1c97af4" />

**Environment**
- LibreBooking version: **4.2.0**
- Installation type: Barebones
- Web server: Apache
- PHP version: **PHP 8.3.6** 
- MariaDB version: **10.11.14**
- Tested on two separate instances with the same result
- Both instances are fresh barebones installs

**Additional context**
- I have confirmed the config file is being read (other settings work).
- Template cache was cleared after the change.
- The setting worked as expected in earlier versions of Booked Scheduler but appears broken or ignored in v4.2.0.

**Possible solution / Workaround**  
None found so far. Setting it to `'current'` or `'none'` changes the behavior, but `'future'` does not enforce the restriction.

belcirelk:
Hi, thanks for the bug report. 

I have tested on my system and I can't book in the past when 'future' is set. Are you sure your regular user is not a resource admin ? I have the latest development version.

Anthony-cloud-1:
[...]

PR #1228, opened 2026-03-25
docs(config): document admin exemption from start.time.constraint

JohnVillalovos:
Clarify that application, group, resource, and schedule admins are exempt from the reservation.start.time.constraint setting.

Related: #1225


belcirelk (approved):
It's clear answer to a frequent confusion.

belcirelk:
Very important to document these things when we can, sometimes it's seen like a bug while it's the normal behavior.

PR #1241, opened 2026-03-27
fix(i18n): add missing embedded_datetime date format to en_gb locale

JohnVillalovos:
The embedded_datetime key was missing from en_gb.php, causing PDF exports, embedded calendar views, and print attribute views to display dates in US format (m/d with AM/PM) instead of the expected d/m with 24-hour time. Since fr_fr and 22 other locales inherit dates from en_gb, this affected all of them.

Issue solved by: labmecanicatec <labmecanicatec@udistrital.edu.co>

### DIFF
--- a/lang/en_gb.php
+++ b/lang/en_gb.php
@@ -31,6 +31,8 @@ class en_gb extends en_us
         $dates['momentjs_datetime'] = 'D/M/YY H:mm';
         $dates['calendar_time'] = 'H:mm';
         $dates['calendar_dates'] = 'd M';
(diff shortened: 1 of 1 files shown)


labmecanicatec (approved):
I would put `$dates['embedded_datetime'] = 'd/m/Y H:i';` between `calendar_dates' and `report_date` for greater consistency with `en_us.php`

JohnVillalovos:
> I would put `$dates['embedded_datetime'] = 'd/m/Y H:i';` between `calendar_dates' and `report_date`for greater consistency with`en_us.php`

Done

PR #1248, opened 2026-03-28
fix: correct check-in status display in resource details popup

### LINKED ISSUE #1240: [Bug] The resource info popup has wrong status for checkin
### Acknowledgements

### Describe the bug

The info popup displays an incorrect check-in status for resources.

### To Reproduce

In the schedule view, hover the mouse over a resource name to trigger the popup. Resource info popup (triggered by hovering over the resource name in the schedule) is displaying the wrong check-in status when checkin is not required.

### Expected behavior

The status should match the resource page status. 

### Screenshots

_No response_

### Additional context

_No response_

### LibreBooking version

e202d44

belcirelk:
The isCheckInEnabled variable was checked with `neq ''` instead of a boolean check, causing the check-in notification to always display regardless of the actual check-in setting.

Closes: #1240

### DIFF
--- a/tpl/Ajax/resourcedetails.tpl
+++ b/tpl/Ajax/resourcedetails.tpl
@@ -135,7 +135,7 @@
                         {translate key='AutoReleaseNotification' args=$autoReleaseMinutes}
                     </div>
                 {/if}
-                {if $isCheckInEnabled neq ''}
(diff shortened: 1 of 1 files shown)


JohnVillalovos (approved):
Thanks @belcirelk 

LGTM

PR #1263, opened 2026-03-30
fix(datepicker): fix Flatpickr 1px calendar width on hidden date pickers

### LINKED ISSUE #1230: Reservation - Repeat - Custom empty
## Describe the bug**
When trying to make a new reservation with option ->Repeat -> Custom 
clicking in the "Repeat On" field opens a small scroll bar without content.

## To Reproduce**
Got to new reservation -> Repeat -> Custom -> Repeat On field 

## Expected behavior**
Dropdown menu should show content/calendar?

LibreBooking version develop - docker.

Image

JohnVillalovos:
Flatpickr's internal setCalendarWidth() measures offsetWidth during
initialization. When a date picker lives inside a hidden ancestor
(d-none, closed Bootstrap collapse, or unopened modal), offsetWidth
returns 0, producing a 1px-wide calendar that renders as an empty
scrollbar.

Initialize Flatpickr eagerly so _flatpickr is always available to
consumers, then recalculate the calendar width once all hidden ancestors
become visible. Visibility is detected via MutationObserver for d-none
class removal, shown.bs.collapse, and shown.bs.modal events. Nested
hidden ancestors are handled recursively.

The width recalculation uses Flatpickr's public set() API:
fp.set('showMonths', fp.config.showMonths) triggers the internal
setCalendarWidth() callback, which re-measures offsetWidth at a point
where the element has its correct layout dimensions.

Closes: #1230

### DIFF
--- a/tpl/Controls/DatePickerSetup.tpl
+++ b/tpl/Controls/DatePickerSetup.tpl
@@ -11,12 +11,6 @@
                 }
             }
 
-            function initFlatpickrOnce(element, config) {
-                if (!element._flatpickr) {
-                    flatpickr(element, config);
-                }
-            }
-
             function initDatePicker_{$ControlId}() {
             const control = document.getElementById("{$ControlId}");
             if (!control) return;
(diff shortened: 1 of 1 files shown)


labmecanicatec (approved):
It works correctly.

PR #1290, opened 2026-04-02
refactor(dashboard-tooltip): show reservation tooltip only on .reserv…

labmecanicatec:
…ationTitle

Updated dashboard_reservation.tpl and dashboard.js so that the tooltip triggers only on the field with class .reservationTitle. This prevents the reservation popup from appearing in the entire row.

### DIFF
--- a/Web/scripts/dashboard.js
+++ b/Web/scripts/dashboard.js
@@ -16,37 +16,13 @@ function Dashboard(opts) {
 
     var reservations = $('.reservation');
 
-    function attachReservationTooltip(reservations, options) {
-      reservations.on('mouseenter', function () {
-        var me = $(this);
-        var refNum = me.attr('id');
-
-        me.attr('data-bs-toggle', 'tooltip').tooltip('show');
-
-        $.ajax({
-          url: options.summaryPopupUrl,
-          data: { id: refNum },
-        })
-          .done(function (html) {
-            me.attr('data-bs-original-title', html).tooltip('show');
-          })
-          .fail(function (xhr, status, error) {
-            me.attr('data-bs-original-title', status + ': ' + error).tooltip('show');
-          });
-      });
-
-      reservations.on('mouseleave', function () {
-        $(this).tooltip('hide');
-      });
-    }
-
     $(document).ready(function () {
-      var reservations = $('.reservation');
-      var options = {
-        summaryPopupUrl: 'ajax/respopup.php',
-      };
+      var reservationTitles = $('.reservationTitle');
 
-      attachReservationTooltip(reservations, options);
+      reservationTitles.each(function () {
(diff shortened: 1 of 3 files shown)


JohnVillalovos:
The 'scope' can not have spaces.  So `refactor(dashboard_tooltip): ...` should work

Web/scripts/dashboard.js
```diff
@@ -16,10 +16,10 @@ function Dashboard(opts) {
...
 
-    function attachReservationTooltip(reservations, options) {
-      reservations.on('mouseenter', function () {
+    function attachReservationTooltip(reservationTitles, options) {
+      reservationTitles.on('mouseenter', function () {
         var me = $(this);
-        var refNum = me.attr('id');
+        var refNum = me.closest('.reservation').attr('id');
 
         me.attr('data-bs-toggle', 'tooltip').tooltip('show');
```
JohnVillalovos: Code Review by Gemini

**5. Empty Tooltip Shown Before Data Loads**
Calling `.tooltip('show')` immediately on mouseenter (line 24) shows an empty/blank tooltip while waiting for the AJAX request. It would be better to show a loading state or wait until the data is ready.

Web/scripts/dashboard.js
```diff
@@ -28,25 +28,25 @@ function Dashboard(opts) {
           data: { id: refNum },
         })
           .done(function (html) {
-            me.attr('data-bs-original-title', html).tooltip('show');
+            me.attr('data-bs-title', html).tooltip('show');
```
JohnVillalovos: Code Review by Gemini

**4. Async Race Condition with Tooltip Display**
If the user hovers out before the AJAX call completes, the tooltip will still be shown when the AJAX finishes, creating an orphaned tooltip. You should check if the user is still hovering before calling `.tooltip('show')` in the `.done()` callback.

JohnVillalovos (commented):
Code Review by Gemini

I have identified two additional issues during my review.

labmecanicatec:
The approach changes; now reservationPopup.js is loaded.

JohnVillalovos:
An AI comment:

In `dashboard_reservation.tpl:8-9`, the span has `data-bs-custom-class="respopup-tooltip"` and `data-bs-html="true"` as HTML attributes. However, `reservationPopup.js:10-12` sets these same attributes via JS before creating the tooltip:

```
  me.attr('data-bs-toggle', 'tooltip')
    .attr('data-bs-html', 'true')
    .attr('data-bs-custom-class', 'respopup-tooltip')
    .attr('data-bs-title', html)
    .tooltip('show');
```

  The template attributes are harmless but redundant — they were needed by the old inline code but attachReservationPopup sets them itself. Removing them from the template would be cleaner.

PR #1291, opened 2026-04-03
refactor(manage-blackouts): tooltips to appear only on .reservationTitle

labmecanicatec:
Updated manage_blackouts_response.tpl and blackouts.js so that the tooltip triggers only on the field with class .reservationTitle. Prevents the reservation tooltip from appearing in the entire row.


JohnVillalovos (approved):
Thanks @labmecanicatec 

LGTM

PR #1325, opened 2026-04-09
fix(ui): align schedule toolbar icons and update icon images

labmecanicatec:
- Normalize icon size to 24px using a shared class
- Fix vertical alignment by applying flex utilities to anchor elements
- Remove incorrect sizing applied to parent <a> elements
- Update schedule view icons (table, tall, wide, week)


JohnVillalovos (approved):
Thanks @labmecanicatec 

LGTM

Image

Issue #444, opened 2024-10-17
Unauthenticated users cannot see reservation details


gitthismoney:
Hello,

We are testing the latest release and I ran in to an issue.

We typically allow unauthenticated users to browse the schedule with full reservation details so we can display Resource Calendar on a tv. 

On the latest release, unauthenticated users are not able to see full reservation details. They only see the booked times, but no title, owner or desc. I attached a screenshot as example. 

If I highlight one of the bookings, it does tell me which room is booked... but I also receive a red message: 
Can't see reservation details because you don't have permissions to any of the resources in this reservation

my config.php 
$conf['settings']['privacy']['view.schedules'] = 'true';
$conf['settings']['privacy']['view.reservations'] = 'true';
$conf['settings']['privacy']['hide.user.details'] = 'true';
$conf['settings']['privacy']['hide.reservation.details'] = 'false';
$conf['settings']['privacy']['allow.guest.reservations'] = 'false';

![Screen Shot 2024-10-17 at 1 50 05 PM](https://github.com/user-attachments/assets/a0526ddd-9618-45ff-a65e-1f1768c9fc5e)

Wondering if i've missed something? 

Thank you,

gitthismoney:
Update, i tested an earlier release librebooking-2.8.6.zip

This release is not affected by the issue

treborin:
> Update, i tested an earlier release librebooking-2.8.6.zip
> 
> This release is not affected by the issue

Pls., @gitthismoney , close this issue

tkasch:
Please don't close this issue without fixing it. Similar to @gitthismoney, we rely on schedule details to be visible for unauthenticated users as well at our site. Since this has worked perfectly for us until version 2.8.6, I'd consider this issue a regression.

I stumbled across this problem back in September 2024 when trying to update (a copy of) our production site in order to do a quick test if LibreBooking supports PHP >= 8.2. Unfortunately, I haven't had the resources to dig into this yet.

For me, such a fundamental change of functionality in a sub-micro-version update (from 2.8.6 to 2.8.6.2) came quite unexpected and IMHO would have justified a minor version bump (at least) as well as an explicit hint in the release notes.

PR #1331, opened 2026-04-09
fix: allow anonymous users to view reservation details when view.reservations=true

nkrings-nettersheim:
## Problem

When `privacy.view.reservations` is set to `true` in config.php,
anonymous (unauthenticated) users were still unable to see reservation
details. They could see that a slot was booked, but no title, name or
description. Clicking on a reservation showed the error:

> "Can't see reservation details because you don't have permissions
> to any of the resources in this reservation"

Related issue: https://github.com/LibreBooking/app/issues/444

## Root Cause

In three places, the code calls `UserResourcePermissions()` which
queries the database for resource permissions by UserId. Anonymous
users have `UserId=0`, so this query always returns an empty array,
causing all details to be hidden regardless of the privacy settings.

## Fix

For anonymous users when `view.reservations=true`:
- Skip the DB permission lookup
- Grant access directly to the resource IDs of the requested reservation
- Logged-in users are completely unaffected

## Files Changed

- `lib/Application/Schedule/SlotLabelFactory.php` – slot labels in schedule view
- `Pages/Ajax/ReservationPopupPage.php` – reservation popup
- `lib/Application/Reservation/ReservationComponentBinder.php` – reservation detail page

## Testing

1. Set in config.php:
   - `privacy.view.schedules = true`
   - `privacy.view.reservations = true`
2. Open schedule as anonymous user (not logged in)
3. Reservation details (title, name etc.) are now visible in:
   - Schedule slot labels ✅
   - Reservation popup ✅
   - Reservation detail page ✅

### DIFF
--- a/lib/Application/Reservation/ReservationComponentBinder.php
+++ b/lib/Application/Reservation/ReservationComponentBinder.php
@@ -389,8 +389,8 @@ class ReservationDetailsBinder implements IReservationComponentBinder
     {
         $userSession = ServiceLocator::GetServer()->GetUserSession();
 
-        // ÄNDERUNG: Anonymer User mit view.reservations=true bekommt die Ressourcen
-        // der aktuellen Reservierung direkt – keine DB-Berechtigungsabfrage nötig
+        // For anonymous users with view.reservations=true, use the resource IDs
+        // of the current reservation directly - no DB permission lookup needed
         $allowGuestView = Configuration::Instance()->GetKey(
             ConfigKeys::PRIVACY_VIEW_RESERVATIONS,
             new BooleanConverter()
@@ -405,7 +405,7 @@ class ReservationDetailsBinder implements IReservationComponentBinder
             return;
         }
 
-        // Eingeloggte User: bisherige Logik unverändert
+        // Logged-in users: keep existing logic unchanged
         $resourceRepo = new ResourceRepository();
         $resourceIds = [];
 
--- a/lib/Application/Reservation/ReservationComponentBinder.php
+++ b/lib/Application/Reservation/ReservationComponentBinder.php
@@ -381,10 +381,31 @@ class ReservationDetailsBinder implements IReservationComponentBinder
 
     /**
      * Gets the resources the user has permissions (full access and view only permissions)
-     * This is used to block a user from seeing reservation details if he has no permissions to it's resources
+     * This is used to block a user from seeing reservation details if he has no permissions to it's resources.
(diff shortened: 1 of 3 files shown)


lib/Application/Reservation/ReservationComponentBinder.php
```diff
@@ -381,22 +381,43 @@ private function SetAutoReleaseMinutes()
...
-     * This is used to block a user from seeing reservation details if he has no permissions to it's resources
+     * This is used to block a user from seeing reservation details if he has no permissions to it's resources.
+     * For anonymous users: if view.reservations=true, grant access to all resources
+     * of the current reservation directly (no DB permission lookup needed).
      */
     private function UserResourcePermissions($userId)
     {
+        $userSession = ServiceLocator::GetServer()->GetUserSession();
+
+        // ÄNDERUNG: Anonymer User mit view.reservations=true bekommt die Ressourcen
```
JohnVillalovos: Can the comments be translated to English please?

JohnVillalovos (changes requested):
Thanks @nkrings-nettersheim 

I'll will try to fully review this no later than the end of the weekend. Hopefully before that.

lib/Application/Reservation/ReservationComponentBinder.php
nkrings-nettersheim: Now you should see the comments in English :-)

JohnVillalovos:
@nkrings-nettersheim Thanks for the update. Can you look at the Copilot comments and respond. Perfectly reasonable to disagree with them.

Also my first thought on this is that maybe we should add a new config setting for this behavior. As changing the behavior of the current setting could potentially be considered a security issue. If some people were depending on reservations not being visible and then they become visible without any changes to their config. What do you think?

belcirelk:
> @nkrings-nettersheim Thanks for the update. Can you look at the Copilot comments and respond. Perfectly reasonable to disagree with them.
> 
> Also my first thought on this is that maybe we should add a new config setting for this behavior. As changing the behavior of the current setting could potentially be considered a security issue. If some people were depending on reservations not being visible and then they become visible without any changes to their config. What do you think?

I don't know if you already solved that,  but apparently it was working in old versions. But that make sense to have a setting specifically for guests to allow or not the view of details. Not everyone would want to display that.

Tovaryon:
Greetings

What is the current status on this? We'd like to switch from the old Booked Schedule to LibreBooking, but this bug is currently preventing us from doing so, and we urgently need this feature.

PR #1337, opened 2026-04-11
refactor(reservation): build PDF config in PHP

JohnVillalovos:
Move the ReservationPdfConfig data structure out of the Smarty template and build it as a PHP array before encoding it to JSON. This keeps the template responsible only for emitting the already-structured config and avoids hand-building a large JavaScript object with mixed escaping rules.

Delegate PDF date formatting to the existing Smarty date formatter so the generated config stays aligned with template date behavior. Keep the temporary SmartyVar bridge private and explicit while the reservation page continues to receive most view data through existing Smarty assignments.

Use JSON_HEX_* encoding flags when emitting the config to keep the inline script safe for values containing quotes, ampersands, or script-like text.

Assisted-by: OpenAI Codex (GPT-5)


labmecanicatec (approved):
The PDFs are generated correctly

PR #1358, opened 2026-04-23
fix: modify .htaccess redirect rule / remove R=301 from htaccess redirect to prevent port leakage behind reverse proxy

wiHiam:
Change redirect rule to use [L] flag instead of [R=301].

The R=301 flag causes Apache to build an absolute URL including the port, breaking setups behind a reverse proxy.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -7,7 +7,7 @@ RewriteEngine On
 
 # Redirect requests that are not already under /Web to /Web.
 RewriteCond %{REQUEST_URI} !^/Web(/|$)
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
@wiHiam Thanks for the PR. After a LOT of work by @colisee this is not needed. Changes were made in the docker repository to fix the issue.

New docker images should be working.

Based on those changes, I am going to close this PR. Thanks again.

colisee:
@wiHiam , for more details, please have a look at issue 200 inside the repo librebooking/docker. Even if you don’t use containers, the solution should work.

PR #1381, opened 2026-05-01
fix(schedule): replace html2canvas with html2canvas-pro to support co…

labmecanicatec:
…lor-mix()

- The original html2canvas library does not support modern CSS Color Level 4 functions
- Functions like color-mix(), oklch(), and lab() cause rendering errors.
- html2canvas-pro is an active fork that properly handles these CSS color functions
- The API remains 100% compatible, requiring no changes to existing implementation code
- Fixes table printing issues where colors defined with color-mix() were not displaying
- Also enables future use of other modern CSS color functions without breaking changes


JohnVillalovos (approved):
Thanks @labmecanicatec 

LGTM

PR #1393, opened 2026-05-06
refactor(reservation): UI improvements

labmecanicatec:
-The Reference Number moves below the title.
- Prevents double lines from being displayed when no custom attributes exist.
- tpl/Ajax/reservation/reservation_attributes.tpl is simplified; a new row is no longer needed every three attributes.


JohnVillalovos (approved):
Thanks @labmecanicatec 

LGTM

PR #1394, opened 2026-05-06
fix(reservation.label): avoid null replacement in attribute label for…

labmecanicatec:
…matting

- When using a label format like: 'reservation.label' => '{att120}' and the attribute with that ID does not exist, GetAttributeValue()
returns null. This caused the following deprecated warning in PHP:    str_replace(): Passing null to parameter 2 ($replace)
    of type array|string is deprecated
SlotLabelFactory.php (line 119)
- The fix ensures null values are converted to an empty string before
calling str_replace, preventing the warning and keeping label rendering stable.


JohnVillalovos (approved):
Thanks @labmecanicatec 

LGTM

PR #1395, opened 2026-05-08
fix(DeprecatedPHP8.5): Transforming null values ​​to an empty string

labmecanicatec:
- Null values ​​are handled to improve compatibility with PHP 8.5 and avoid "Deprecated" messages

Assisted-by: Copilot:GPT-4.5


JohnVillalovos (approved):
Thanks @labmecanicatec 

LGTM

PR #1397, opened 2026-05-08
fix: Grammar and spelling errors fixes for german localization

AurelioSilver:
I fixed a couple of spelling and grammar errors and improved the german translation.


JohnVillalovos (approved):
Thanks @AurelioSilver 

LGTM, as someone who doesn't know any German 😄

PR #1428, opened 2026-05-26
fix(auth): block unsafe external redirect targets

JohnVillalovos:
Sanitize redirect destinations before issuing Location headers so login, SSO callback, logout, and resume flows cannot forward users to attacker- controlled URLs.

Allow only internal relative redirects and same-origin absolute HTTP(S) URLs. Reject protocol-relative, malformed, and non-HTTP scheme targets by falling back to a safe internal page.

Add shared redirect sanitizer coverage and login presenter regression tests for external redirect payloads.

Reported-by: Lorenzo Russo <lorenzorusso2911@gmail.com>
Assisted-by: Codex:GPT-5

### DIFF
--- a/Pages/Page.php
+++ b/Pages/Page.php
@@ -160,23 +160,26 @@ abstract class Page implements IPage
 
     public function Redirect($url)
     {
-        if (!BookedStringHelper::StartsWith($url, $this->path)) {
-            $url = $this->path . $url;
-        }
-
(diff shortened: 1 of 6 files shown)


tienneR:
Hi John, is it possible to ask for a CVE?

JohnVillalovos:
> Hi John, is it possible to ask for a CVE?

We are just a few volunteers working on this project. So we don't have any procedures setup for that.

PR #1440, opened 2026-06-03
schedule.tpl fix for Smarty Compiler: Syntax error

amikoreh:
Cause: Smarty cannot parse the PHP enum expression ScheduleStyle::Standard->value (the -> inside {if ...}), so the template failed to compile.
Fix applied: Replaced the enum expressions with their integer values (0..3) in [schedule.tpl:100-140]


JohnVillalovos:
I don't believe this is needed. You should just need to update your version of Smarty.

JohnVillalovos:
See https://github.com/smarty-php/smarty/releases/tag/v5.8.0

And https://github.com/smarty-php/smarty/pull/1171

amikoreh:
@JohnVillalovos You are correct. Smarty update fixed it. Thanks for the super quick response and all your work on LibreBooking!

JohnVillalovos:
> @JohnVillalovos You are correct. Smarty update fixed it. Thanks for the super quick response and all your work on LibreBooking!

You're welcome. Glad to hear it worked.

PR #1469, opened 2026-06-09
feat(ics): add global config switch to enable/disable ICS/Atom feeds

JohnVillalovos:
NOTE: @lucs7 is the original author of this. Splitting PR #1425 into multiple commits to make reviewing easier.

Add a new `ics.enabled` boolean config key (default: true) that acts as a master on/off switch for ICS/Atom calendar subscription feeds.

When set to false:
- CalendarSubscriptionValidator returns false immediately, resulting in
  a 404 for any subscription request, with a debug log entry
- CalendarSubscriptionDetails::IsEnabled() short-circuits to false
- Subscription links and the bulk-edit "Turn On Subscription" control
  are hidden in the Manage Resources and Manage Schedules admin UIs
- SchedulePageBuilder suppresses the subscription URL, hiding
  subscription links in the end-user schedule view

Also fix indentation of the QR code link block in manage_resources.tpl.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -438,6 +438,9 @@ LB_TABLET_VIEW_AUTO_SUGGEST_EMAILS=false
 # ICS Settings
 ##############
 
+# Enable or disable ICS calendar subscription feeds entirely (true/false)
+LB_ICS_ENABLED=true
+
 # Subscription key secret used for ICS calendar feeds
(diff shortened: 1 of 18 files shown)


JohnVillalovos:
@lucs7 I think this is ready, but would like to get your sign-off before merging.

PR #1486, opened 2026-06-14
fix(reservations): prevent false success for no-op updates

### LINKED ISSUE #1430: [Bug] Delete/modify reservations misbehavior while being resource administrator
### Acknowledgements

### Describe the bug

I'm not able to delete or modify currently active/past reservations despite being resource administrator, although UI says that reservation been successfully deleted/modified.

### To Reproduce

1. Create a group with resource administrator rights, create a resource, grant rights for this resource to created group
2. Add your user to created group only
3. From behalf of this user create a reservation, that starts soon
4. Wait till the time of reservation starts
5. Try to delete/modify time of the reservation
6. UI says that reservation was successfully deleted
7. Observe the remaining reservation. 

### Expected behavior

I expect active reservations (reservations which time already being started) to be deletable by resource administrators

### Screenshots

_No response_

### Additional context

_No response_

### LibreBooking version

5.0.2

JohnVillalovos:
When start time constraints filter all selected reservation instances out of an update or delete operation, validation still passed and the UI reported success even though nothing was changed.

Add a validation rule for update and delete operations that fails when the selected scope contains no affected reservation instances. This preserves the current restriction for active or past reservations while showing the existing start-time constraint error instead of a misleading success message.

Note: this does not grant resource administrators the ability to act on active reservations — that requires extending the Instances() bypass (currently limited to application admins) to cover resource admins as well.

Closes: #1430
Assisted-by: Codex:GPT-5

### DIFF
--- a/lib/Application/Reservation/Validation/PreReservationFactory.php
+++ b/lib/Application/Reservation/Validation/PreReservationFactory.php
@@ -149,6 +149,7 @@ class PreReservationFactory implements IPreReservationFactory
 
     private function CreateUpdateService(ReservationValidationRuleProcessor $ruleProcessor, UserSession $userSession)
     {
+        $ruleProcessor->AddRule(new ReservationInstancesRule());
         if (Configuration::Instance()->GetKey(ConfigKeys::RESERVATION_UPDATES_REQUIRE_APPROVAL, new BooleanConverter())) {
             $ruleProcessor->AddRule(new AdminExcludedRule(new RequiresApprovalRule(PluginManager::Instance()->LoadAuthorization()), $userSession, $this->userRepository));
         }
@@ -160,6 +161,7 @@ class PreReservationFactory implements IPreReservationFactory
 
     private function CreateDeleteService(ReservationValidationRuleProcessor $ruleProcessor, UserSession $userSession)
     {
+        $ruleProcessor->AddRule(new ReservationInstancesRule());
         $ruleProcessor->AddRule(new AdminExcludedRule(new ResourceMinimumNoticeRuleDelete($userSession), $userSession, $this->userRepository));
         $ruleProcessor->AddRule(new AdminExcludedRule(new CurrentUserIsReservationUserRule($userSession), $userSession, $this->userRepository));
         return new DeleteReservationValidationService($ruleProcessor);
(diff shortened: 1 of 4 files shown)


JohnVillalovos:
Intentionally did not make a change to grant resource administrators the ability to modify past reservations.

Fixed the false-success behavior in this path, but intentionally did not expand resource administrator permissions to allow modifying or deleting past/already started reservations.

The current behavior is tied to `reservation.start.time.constraint`. With the
default `future` setting, reservations that have already started are not
eligible for normal update/delete operations. Resource administrators were
bypassing the start-time validation rule, but the later series-scope filtering
still selected zero affected reservation instances. That is why the UI could
report success even though nothing changed.

The fix keeps the existing policy intact and makes the result honest: when the
configured constraint leaves no reservation instances eligible to change, the
operation now fails with the existing start-time constraint message instead of
reporting success.

I do not think we should change resource administrator behavior as part of this
bug fix. Allowing resource administrators to modify active or historical
reservations is a larger product/security decision with several side effects:

- It changes the meaning of "resource administrator" from managing resource
  availability and future bookings to being able to alter the historical record
  for that resource.
- It can affect reports, utilization history, quota enforcement, credits, and
  any external billing/accounting workflows that depend on reservation history
  being stable after the configured constraint boundary.
- It complicates recurring reservations. We would need precise behavior for
  "this instance", "future instances", and "full series" when some instances
  are already started or completed and others are still future reservations.
- It complicates multi-resource reservations. If a reservation uses multiple
  resources and the user administers only one of them, changing or deleting the
  entire reservation could affect resources outside that administrator's
  authority.
- It raises notification and calendar-sync questions. Owners, participants,
  invitees, resource admins, and external calendars may already have been
  notified about the original reservation. Changing past or active reservations
  can make those downstream records inconsistent unless we define exactly what
  gets sent and when.
- It interacts with check-in/check-out state. A reservation that is checked in,
  checked out, auto-released, or missing check-out may need different rules than
  a normal future reservation.
- It can surprise deployments that intentionally use
  `reservation.start.time.constraint = future` to prevent non-application-admin
  users from changing reservations after they begin.
- It may require additional audit logging or a separate permission/configuration
  switch so deployments can choose whether resource administrators are trusted
  to change active/historical reservations.

For those reasons, this change only fixes the misleading success response. If
we want resource administrators to be able to change active or past
reservations, that should be designed and tested separately with explicit rules
for recurring reservations, multi-resource reservations, checked-in/out
reservations, notifications, reporting, and configuration compatibility.

PR #1500, opened 2026-06-21
Refactor Microsoft Auth into a testable class with expanded error handling

barakiva:
## Summary

`Web/microsoft-auth.php` previously handled only the success path of
Microsoft's OAuth 2.0 authorization code flow. Per the Microsoft identity
platform docs, the callback endpoint receives two distinct response shapes:

- Success: `?code=...&state=...`
- Error:   `?error=...&error_description=...`

Error responses were silently ignored — the page would redirect to
`Web` with no `error` header, causing a confusing failure
downstream with no diagnostic information.

## Changes

- Extract callback logic into `MicrosoftOAuthCallback` (lib/Application/Authentication/)
  with a `MicrosoftOAuthCallbackResult` value object representing the two
  outcome cases. This makes the code more testable.
- Handle error responses from Azure: redirect to home and log the error/description
- unit tests covering success, URL encoding, error with/without description,
  error-takes-precedence-over-code, and missing params
- Add `workflow_dispatch` trigger to CI workflows to allow manual runs on forks so contributors can make sure PRs will pass CI before submitting to upstream

## Not addressed in this PR

- `state` parameter verification — the Microsoft docs recommend verifying the
  returned `state` matches the value sent in the authorization request as a
  CSRF mitigation. This would require storing the state server-side before the
  redirect, which is a larger change.

Any feedback on testing & class design would be appreicated.

### DIFF
--- a/Web/microsoft-auth.php
+++ b/Web/microsoft-auth.php
@@ -3,13 +3,17 @@
 define('ROOT_DIR', '../');
 
 require_once(ROOT_DIR . 'lib/Common/namespace.php');
+require_once(ROOT_DIR . 'lib/Application/Authentication/MicrosoftOAuthCallback.php');
 
-//Checks if the user was authenticated by microsoft and redirects to external authentication page
-if (isset($_GET['code'])) {
-    $code = filter_input(INPUT_GET, 'code');
-    header('Location: '.ROOT_DIR.'Web/external-auth.php?type=microsoft&code='.$code);
-    exit;
-} else {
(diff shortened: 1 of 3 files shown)


barakiva:
Applied all code reivew suggestions and unit tests all pass.

JohnVillalovos (approved):
Thanks @barakiva 

LGTM

Probably should do a follow-up to minimize potential log spam from things hitting the endpoint and thus it get logged. But that can be a follow-up.

PR #1529, opened 2026-07-08
refactor(reservations): initialize reservation with rendered $StartDa…

labmecanicatec:
…te and $EndDate

- Replace the use of values ​​read from #BeginDate and #EndDate with dates rendered by Smarty, avoiding reliance on the DOM state during reservation initialization.
- Maintains consistency with Reservation/create.tpl
- Avoids the use of jQuery


JohnVillalovos (approved):
Thanks @labmecanicatec 

Makes sense to me.

Issue #1512, opened 2026-07-02
[Bug] App crashes when I do a reservation in a calendar that start in the future, starting from the dashboard


apezzi75:
### Acknowledgements

- [x] This is a bug and not a question. Questions should be asked in the [Discussions area](https://github.com/LibreBooking/librebooking/discussions).
- [x] This project is maintained 100% by volunteers. I understand that there are no guarantees on when (or if) this will be addressed, and that the most effective way to see this resolved is for me to submit a Pull Request.

### Describe the bug

App freezes when I click on calendar picker. Start Date and End Date fields are blank because the calendar is available starting September 15th. Reservations are possible through other channels, however. When I click on the calendar date to choose a possible date (e.g., September 16th), the application crashes. 

### To Reproduce

1. Goto Dashboard
2. Click on Reservation Button
3. Click on calendar picker for choosing the date

### Expected behavior

The app must not crash

### Screenshots

<img width="1613" height="552" alt="Image" src="https://github.com/user-attachments/assets/d735990a-2af8-4e69-89d5-960fc99562f6" />

### LibreBooking version

5.1 develop

belcirelk:
It looks like a javascript issue. Can you check in your browser console if you have any error message? 

I am not sure to understand the September 15, is this a setting you added?

apezzi75:
> It looks like a javascript issue. Can you check in your browser console if you have any error message?
> 
> I am not sure to understand the September 15, is this a setting you added?

The calendar is set to be available for bookings starting September 15th.

belcirelk:
I have tried to reproduce this, but I can only access the reservation page via calendar, not schedule. So maybe the real bug is that we are not supposed to be able to reserve if the schedule is not available. 

But it's a real bug, the page get stuck in a loop of 500 errors. 

The alternative would be that you block the resources and leave the schedule always available.

belcirelk:
After more testing, here the bug description using Claude Sonnet 5. Let me know @apezzi75 if I missed something. 

**Describe the bug**
When a schedule is configured with a limited Schedule Availability window (e.g., available from Sept 15, 2026 onward). Three different entry points into the reservation flow behave differently when today's date is before the window starts:

- **Schedule view** (direct): blocks all access entirely, showing "This schedule is not yet available," even for dates that fall inside the configured window.
- **Find a Time**: allows selecting and booking any date within the configured window, even though today is before the window starts.
[...]

PR #1534, opened 2026-07-09
fix(reservations): stop browser freeze from unbounded date-advance loop

### LINKED ISSUE #1533: [Bug] Reservation page freezes browser: unbounded date-advance loop in PopulatePeriodDropDown (reservation.js)
### Acknowledgements

### Describe the bug

When the reservation page (`reservation.php`) is opened for a schedule/date combination where the selected day has no reservable periods — most notably a schedule whose availability window begins in the future — the page can lock up the browser tab completely.

**Relationship to #1512:** Issue #1512 reports this freeze as experienced from one specific entry point: clicking Reserve on the dashboard's Resource Availability widget for a schedule that starts in the future. The dashboard arrives at the reservation page without usable start/end dates, and the page then hits the loop described below. This issue is filed separately because the defect is in the reservation page's own JavaScript, not in the dashboard: any path into the reservation page (dashboard links, bookmarks, direct URLs, other pages) can trigger it. Fixing the dashboard link only shrinks the entry surface; fixing this loop resolves the freeze at the root for all entry points.

The problem is in first-party code (`Web/scripts/reservation.js`), not in the bundled Flatpickr library. Three defects compound:

1. **Unbounded day-advance loop.** When the selected day yields zero reservable periods, `PopulatePeriodDropDown` advances the date by one day and re-triggers `change` on itself, with no upper bound and no respect for the picker's `minDate`/`maxDate`:

   ```js
   // Web/scripts/reservation.js:925-929
   if (items.length === 0) {
       var nextDate = new Date(dateElement.val());
       nextDate.setDate(nextDate.getDate() + 1);
       dateElement[0]._flatpickr.setDate(nextDate, true);
       dateElement.trigger('change');
   }
   ```

   Because `layoutCache` is keyed by **weekday** (0–6), after at most 7 iterations every lookup is a cache hit and the loop degenerates into a pure CPU spin — the tab freezes hard.

2. **The re-entry guard never works.** The early exit at `Web/scripts/reservation.js:883` compares `prevDate.getTime() === currDate.getTime()`. The previous value is stored via `elements.beginDate.data['beginPreviousVal'] = ...` (indexing the jQuery `.data` *method* rather than calling it, `reservation.js:793-795`) and read back as `dateElement.data('previousVal')` — a key that is never set. `new Date(undefined)` is Invalid Date, `getTime()` is `NaN`, and `NaN === NaN` is false, so the guard can never stop the loop. A blank date field (`new Date('')`) hits the same `NaN` problem.

3. **Synchronous AJAX per iteration.** `getLayoutItems` (`Web/scripts/reservation.js:866-874`) fetches each uncached day's layout with `$.ajax({ ..., async: false })`, blocking the UI thread on every iteration until the weekday cache fills.

### To Reproduce

**Note: these steps are derived from reading the code; they have not yet been verified on a running instance.** The freeze itself has been observed in the wild (#1512); the steps below are the code-level reconstruction of how it is reached.

1. Create a schedule whose availability begin date is in the future (e.g. opens two months from now); assign a resource to it.
2. Open the dashboard and, in Resource Availability, click the Reserve button (or the resource name) for that resource — or open `reservation.php?rid=<id>` directly.
3. Interact with the Begin/End date pickers and select a date.
4. The tab becomes unresponsive (busy loop) and must be killed.

### Expected behavior

If the selected day has no reservable periods, the page should fail gracefully: stop after a bounded number of skipped days, clamp to the schedule's availability window (`minDate`/`maxDate` are already passed to the pickers from `tpl/Reservation/create.tpl:523-524`), or show a message. It must never lock the browser.

### Screenshots

_No response_

### Additional context

- Suggested fix directions:
  - Bail out of `PopulatePeriodDropDown` when the date is invalid (`isNaN(currDate.getTime())`).
  - Bound the advance loop and clamp to the picker's `minDate`/`maxDate` — jumping directly to `minDate` when the current date precedes the availability window would fix the future-schedule case outright.
  - Fix the previous-value bookkeeping to use `.data('key', value)` / `.data('key')`.
  - Longer term, replace the `async: false` layout fetch with an async call.
- Line numbers refer to current `develop`.

### LibreBooking version

develop, `add7ebfbe`

JohnVillalovos:
When a selected day had no reservable periods, PopulatePeriodDropDown advanced the date one day at a time and re-triggered itself with no upper bound. Because the layout cache is keyed by weekday, after seven iterations every lookup was a cache hit and the loop became a pure CPU spin, freezing the tab. This was reachable whenever the reservation page opened on a date outside the schedule's availability window, for example from the dashboard for a schedule that opens in the future.

- Bound the day-advance to seven skips (layouts repeat weekly) and stop at the picker's maxDate; when the current date is before the schedule opens, jump straight to minDate instead of walking day by day
- Bail out of PopulatePeriodDropDown and AdjustEndDate when a date field is blank or invalid instead of doing arithmetic on Invalid Date values
- Store previous field values with jQuery .data(key, value) instead of assigning properties onto the .data method itself; the date-unchanged guard read a 'previousVal' key that was never written, so it could never stop a change cascade
- Parse dates with dateHelper.parseYMDDate so weekday-keyed caches use the local weekday; new Date('Y-m-d') parses as UTC and shifted the weekday in timezones west of UTC
- Treat a failed layout fetch as an empty day instead of letting $.map throw on undefined and abort the remaining change handlers

Closes: #1533
Related: #1512
Assisted-by: Claude:claude-fable-5

### DIFF
--- a/Web/scripts/reservation.js
+++ b/Web/scripts/reservation.js
@@ -617,12 +617,16 @@ function Reservation(opts) {
   };
 
   var AdjustEndDate = function () {
-    var firstDate = new Date(elements.beginDate.data['beginPreviousVal'] + 'T' + elements.beginTime.val());
+    var firstDate = new Date(elements.beginDate.data('previousVal') + 'T' + elements.beginTime.val());
     var secondDate = new Date(elements.beginDate.val() + 'T' + elements.beginTime.val());
+    var currentEndDate = new Date(elements.endDate.val() + 'T' + elements.endTime.val());
+
+    // If any date or time field is blank then no adjustment can be made
+    if (isNaN(firstDate.getTime()) || isNaN(secondDate.getTime()) || isNaN(currentEndDate.getTime())) {
+      return;
+    }
 
     var diffDays = (secondDate.getTime() - firstDate.getTime()) / oneDay;
-
-    var currentEndDate = new Date(elements.endDate.val() + 'T' + elements.endTime.val());
     currentEndDate.setDate(currentEndDate.getDate() + diffDays);
 
     elements.endDate[0]._flatpickr.setDate(currentEndDate, true);
@@ -790,9 +794,9 @@ function Reservation(opts) {
     periodsCache['end'] = [];
     var layoutCache = [];
 
-    elements.beginDate.data['beginPreviousVal'] = elements.beginDate.val();
-    elements.endDate.data['endPreviousVal'] = elements.endDate.val();
-    elements.beginTime.data['beginTimePreviousVal'] = elements.beginTime.val();
+    elements.beginDate.data('previousVal', elements.beginDate.val());
+    elements.endDate.data('previousVal', elements.endDate.val());
+    elements.beginTime.data('previousVal', elements.beginTime.val());
 
     function BeginDateChanged() {
       PopulatePeriodDropDown(elements.beginDate, elements.beginTime, 'begin');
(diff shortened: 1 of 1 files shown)


JohnVillalovos:
I tested this locally and before this PR I was able to reproduce issue #1512  After this commit it stopped the hanging.

labmecanicatec (approved):
It works properly

JohnVillalovos:
> It works properly

Thanks for the review!

PR #1551, opened 2026-07-13
fix(email): render resource description/notes as rich text in confirm…

belcirelk:
…ation email

Resource description and notes fields support HTML when edited, but the reservation confirmation email was rendering them through |escape, which escaped that formatting.

Switch to |sanitize_rich_text|url2link so the email output matches what was actually entered, consistent with how these fields are rendered elsewhere in the app. sanitize_rich_text still strips unsafe markup, so this isn't reintroducing an XSS risk.


JohnVillalovos (approved):
Thanks @belcirelk 

LGTM

Issue #1559, opened 2026-07-16
[Bug] After login on a specific page, should return to that page


florian-enercoop:
### Acknowledgements

- [x] This is a bug and not a question. Questions should be asked in the [Discussions area](https://github.com/LibreBooking/librebooking/discussions).
- [x] This project is maintained 100% by volunteers. I understand that there are no guarantees on when (or if) this will be addressed, and that the most effective way to see this resolved is for me to submit a Pull Request.

### Describe the bug

The current handling when accessing URL needing authentication is unclear (message is vague) and doesn't redirect properly, making it harder to direct access a single page without many clicks.

Below an example with reservation page.

### To Reproduce

- go to a page needing authentication to be visible without being logged in like Web/reservation.php?rn=6a58ce2936774522386700
- "Unknown Error" sent back 
- click on log in to authenticate
- being redirected to home URL
- go again to Web/reservation.php?rn=6a58ce2936774522386700
- page visible

### Expected behavior

- go to a page without being logged in like Web/reservation.php?rn=6a58ce2936774522386700
- text saying "you need to be authenticated" to see this page
- click on log in to authenticate 
- being redirected to Web/reservation.php?rn=6a58ce2936774522386700

### Additional context

using CAS for authentication

### LibreBooking version

5.1.0

belcirelk:
Just tested on 5.2. I don't have the error message, but I am redirected to the login page. 

Using username/password the expected page loads. However using "Sign in with..." Loads the dashboard instead. So I confirm the bug.

belcirelk:
@florian-enercoop which sign in method you have configured that doesn't work? I see that you wrote CAS, do you mean plugins/Authentication/cas?

florian-enercoop:
> [@florian-enercoop](https://github.com/florian-enercoop) which sign in method you have configured that doesn't work? I see that you wrote CAS, do you mean plugins/Authentication/cas?

Yes exactly, I use plugins/Authentication/cas.

PR #1568, opened 2026-07-18
fix(auth): preserve redirect target through Microsoft OAuth login

belcirelk:
The "Sign in with Microsoft" link never carried the original ?redirect= query param through the OAuth round trip, so users who followed a deep link into a protected page (e.g.
index.php?redirect=/Web/reservation.php?rn=...) were sent to their default homepage after authenticating with Microsoft instead of the page they originally requested.

- LoginPresenter::GetMicrosoftUrl() now accepts an optional $state parameter and includes it as the OAuth `state` param on the Microsoft authorize URL. PageLoad() passes the page's resume URL into it via GetResumeUrl().
- MicrosoftOAuthCallback::handle() now reads `state` back from Microsoft's callback params (validated as a string, consistent with the existing code/error handling) and forwards it to external-auth.php as redirect=, which ExternalAuthLoginPage::GetResumeUrl() already reads via QueryStringKeys::REDIRECT.

The forwarded value still passes through
RedirectUrlSanitizer::Sanitize() in LoginRedirector::Redirect() before use, same as the plain-login ?redirect= flow, so open-redirect protection is unchanged.

Refs: #1559
Assisted-by: Claude:claude-sonnet-4-6

### DIFF
--- a/Presenters/LoginPresenter.php
+++ b/Presenters/LoginPresenter.php
@@ -138,7 +138,7 @@ class LoginPresenter
         $oauth2Enabled    = Configuration::Instance()->GetKey(ConfigKeys::AUTHENTICATION_OAUTH2_LOGIN_ENABLED, new BooleanConverter());
 
         $this->_page->SetGoogleUrl($googleEnabled ? $this->GetGoogleUrl() : null);
-        $this->_page->SetMicrosoftUrl($microsoftEnabled ? $this->GetMicrosoftUrl() : null);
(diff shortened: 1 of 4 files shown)


JohnVillalovos:
@belcirelk Shouldn't this close the issue and not just reference it?

belcirelk:
I think the same would occur with Sign in with Google and possibly others. But since I don't use them, I can't test.

belcirelk:
I tested with Google and it's the same problem. But I will follow up for it in another PR.

JohnVillalovos (approved):
Thanks @belcirelk 

LGTM

Note: I squashed it down to one commit and added three more tests.

PR #1584, opened 2026-07-20
feat(calendar): enhance iCalendar export functionality to use sabre

JohnVillalovos:
removes template file for ical


JohnVillalovos:
@lucs7 Can you look at this PR? It is the first commit from your PR. Feel free to make changes on your PR and I can update this one. Or if you want create a new PR with just this one commit.

Thanks!

lucs7:
Thanks for pulling this out but the 3 other commits on top that address exactly what Copilot flagged here: removing the double-escaping (manual pre-escaping + Sabre's own escaping were stacking), switching ExtraIcalLines parsing to Sabre\VObject\Reader so parameters and nested components round-trip correctly, and adding test coverage for all of it. I'll push the full branch so you can pull in the complete set rather than just this single commit.

PR #1601, opened 2026-07-24
docs(README): update demo side to loader

lucs7:
Add a landing page that shows a loading spinner while the free-tier
Fly.io demo instance wakes from being idle, then redirects once it
responds. This avoids sending visitors straight to a cold-starting
instance with no feedback.

Implement it as a Sphinx page (docs/source/demo.rst) rendered by the
existing Furo-themed ReadTheDocs build, rather than a standalone
GitHub Pages page: GitHub Pages isn't configured for this repository,
and its "deploy from branch" source only supports the repo root or
/docs as a whole, not an arbitrary subdirectory, so a bare static
file wouldn't have been reachable at a sensible URL. The page is
styled with Furo's CSS variables so it follows the site's light/dark
toggle, and AbortSignal.timeout is feature-detected before use so
older browsers degrade gracefully instead of throwing.

Point the README's "Try the demo" link at the built ReadTheDocs page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ significantly and diverged from the original project.
 
 A live demo instance of LibreBooking is available for testing:
 
-[Try the demo](https://librebooking-demo.fly.dev/Web/)
+[Try the demo](https://librebooking.readthedocs.io/en/latest/demo.html)
 
 | Role  | Username | Password    |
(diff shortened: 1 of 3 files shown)


JohnVillalovos (approved):
Thanks @lucs7 

LGTM

PR #1617, opened 2026-07-30
fix(dashboard): handle schedule availability windows for resource dis…

labmecanicatec:
…play

Add schedule availability window evaluation to the resource availability dashboard widget. Resources are now filtered based on their schedule's availability date range (start_date/end_date).

Behavior changes:

- PAST schedules (now >= end_date): Resources are hidden from dashboard
- ACTIVE schedules (start_date <= now < end_date): Resources display normally with current reservations/availability
- FUTURE schedules (now < start_date): Resources appear in "Unavailable" section showing the availability start date (without time component) and with pre-filled START_DATE in reservation links
- Update resource links in the Resource Availability  dashboard to open the corresponding schedule  filtered by the selected resource instead of the  Reserve page. This allows users to see when  resources become available and start making
  reservations from their scheduled availability date.

This allows users to see when resources will become available and to start making reservations from their scheduled availability date.

Relates: #1512
Assisted-by: Claude:claude-haiku-4.5

### DIFF
--- a/Presenters/Dashboard/ResourceAvailabilityControlPresenter.php
+++ b/Presenters/Dashboard/ResourceAvailabilityControlPresenter.php
@@ -42,6 +42,13 @@ class ResourceAvailabilityControlPresenter
         $resources = $this->resourceService->GetAllResources(false, $user);
         $reservations = $this->GetReservations($this->reservationViewRepository->GetReservations($now, $now->AddDays(30)));
 
+        // Build schedule lookup table
+        $schedules = $this->scheduleRepository->GetAll();
+        $schedulesById = [];
+        foreach ($schedules as $schedule) {
+            $schedulesById[$schedule->GetId()] = $schedule;
+        }
+
         $available = [];
         $unavailable = [];
         $allday = [];
@@ -50,6 +57,38 @@ class ResourceAvailabilityControlPresenter
             if ($resource->StatusId == ResourceStatus::HIDDEN) {
                 continue;
             }
+
+            // Check schedule availability
+            if (isset($schedulesById[$resource->ScheduleId])) {
+                $schedule = $schedulesById[$resource->ScheduleId];
+                if ($schedule->HasAvailability()) {
+                    $begin = $schedule->GetAvailabilityBegin();
+                    $end = $schedule->GetAvailabilityEnd();
+
+                    // If schedule is past, don't show this resource
+                    if ($now->GreaterThanOrEqual($end)) {
+                        continue;
+                    }
+
+                    // If schedule is future, add to unavailable with availability begin date
+                    if ($now->LessThan($begin)) {
+                        $futureReservation = new ReservationItemView(
+                            startDate: $begin,
+                            endDate: $begin,
+                            resourceName: $resource->GetName(),
+                            resourceId: $resource->GetId()
+                        );
+                        $futureReservation->ResourceColor = $resource->GetColor();
+                        $unavailable[$resource->ScheduleId][] = new UnavailableDashboardItem(
+                            resource: $resource,
+                            currentReservation: $futureReservation,
+                            isFutureScheduleAvailability: true
+                        );
+                        continue;
+                    }
+                }
+            }
+
             $reservation = $this->GetOngoingReservation($resource, $reservations);
 
             if ($reservation != null) {
@@ -60,9 +99,17 @@ class ResourceAvailabilityControlPresenter
                 }
 
                 if (!$reservation->EndDate->DateEquals($now)) {
-                    $allday[$resource->ScheduleId][] = new UnavailableDashboardItem($resource, $lastReservationBeforeOpening);
+                    $allday[$resource->ScheduleId][] = new UnavailableDashboardItem(
+                        resource: $resource,
+                        currentReservation: $lastReservationBeforeOpening,
+                        isFutureScheduleAvailability: false
+                    );
                 } else {
-                    $unavailable[$resource->ScheduleId][] = new UnavailableDashboardItem($resource, $lastReservationBeforeOpening);
+                    $unavailable[$resource->ScheduleId][] = new UnavailableDashboardItem(
+                        resource: $resource,
+                        currentReservation: $lastReservationBeforeOpening,
+                        isFutureScheduleAvailability: false
+                    );
                 }
             } else {
                 $resourceId = $resource->GetId();
@@ -77,7 +124,7 @@ class ResourceAvailabilityControlPresenter
         $this->control->SetAvailable($available);
         $this->control->SetUnavailable($unavailable);
         $this->control->SetUnavailableAllDay($allday);
-        $this->control->SetSchedules($this->scheduleRepository->GetAll());
+        $this->control->SetSchedules($schedules);
     }
 
     /**
--- a/tests/Presenters/Dashboard/ResourceAvailabilityControlPresenterTest.php
+++ b/tests/Presenters/Dashboard/ResourceAvailabilityControlPresenterTest.php
@@ -75,13 +75,67 @@ class ResourceAvailabilityControlPresenterTest extends TestBase
             $this->control->_AvailableNow[1][0]
         );
         $this->assertEquals(
-            new UnavailableDashboardItem($this->unavailableResource, $this->reservationRepo->_Reservations[1]),
+            new UnavailableDashboardItem(
+                resource: $this->unavailableResource,
+                currentReservation: $this->reservationRepo->_Reservations[1],
+                isFutureScheduleAvailability: false
+            ),
             $this->control->_UnavailableNow[1][0]
         );
         $this->assertEquals(
-            new UnavailableDashboardItem($this->unavailableAllDayResource, $this->reservationRepo->_Reservations[2]),
+            new UnavailableDashboardItem(
+                resource: $this->unavailableAllDayResource,
+                currentReservation: $this->reservationRepo->_Reservations[2],
+                isFutureScheduleAvailability: false
+            ),
             $this->control->_UnavailableAllDay[1][0]
         );
+
+        $this->assertFalse($this->control->_UnavailableNow[1][0]->IsFutureScheduleAvailability());
+        $this->assertFalse($this->control->_UnavailableAllDay[1][0]->IsFutureScheduleAvailability());
+    }
+
+    public function testFutureScheduleResourceAppearsAsUnavailable()
+    {
+        Date::_SetNow(Date::Parse('2026-07-29 10:00', 'UTC'));
+
+        $futureSchedule = new FakeSchedule(1);
+        $futureSchedule->SetAvailability(Date::Now()->AddDays(5), Date::Now()->AddDays(35));
+        $this->scheduleRepo->_Schedules = [$futureSchedule];
+
+        $futureResource = new TestResourceDto(1, '1');
+        $this->resourceService->_AllResources = [$futureResource];
+        $this->reservationRepo->_Reservations = [];
+
+        $this->presenter->PageLoad($this->fakeUser);
+
+        $this->assertNotEmpty($this->control->_UnavailableNow);
+        $this->assertNotEmpty($this->control->_UnavailableNow[1]);
+        $unavailableItem = $this->control->_UnavailableNow[1][0];
+        $this->assertInstanceOf('UnavailableDashboardItem', $unavailableItem);
+        $this->assertEquals($futureResource->GetId(), $unavailableItem->ResourceId());
+        $this->assertTrue($unavailableItem->IsFutureScheduleAvailability());
+        // Verify that the availability begin date is set as the end date of the synthetic reservation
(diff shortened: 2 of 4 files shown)


Presenters/Dashboard/ResourceAvailabilityControlPresenter.php
```diff
@@ -50,6 +57,29 @@ public function PageLoad(UserSession $user)
...
+                        continue;
+                    }
+
+                    // If schedule is future, add to unavailable with availability begin date
+                    if ($now->LessThan($begin)) {
+                        $futureReservation = new ReservationItemView(null, $begin, $begin, $resource->GetName(), $resource->GetId());
+                        $futureReservation->ResourceColor = $resource->GetColor();
+                        $unavailable[$resource->ScheduleId][] = new UnavailableDashboardItem($resource, $futureReservation);
+                        continue;
+                    }
```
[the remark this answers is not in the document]
labmecanicatec: Done

tests/Presenters/Dashboard/ResourceAvailabilityControlPresenterTest.php
```diff
@@ -105,7 +105,11 @@ private function PopulateReservations()
 
     private function PopulateSchedules()
     {
-        $this->scheduleRepo->_Schedules = [new FakeSchedule(1), new FakeSchedule(2)];
+        $schedule1 = new FakeSchedule(1);
+        $schedule1->SetAvailability(Date::Now()->AddDays(-30), Date::Now()->AddDays(30));
+        $schedule2 = new FakeSchedule(2);
+        $schedule2->SetAvailability(Date::Now()->AddDays(-30), Date::Now()->AddDays(30));
+        $this->scheduleRepo->_Schedules = [$schedule1, $schedule2];
     }
```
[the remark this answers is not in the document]
labmecanicatec: Okay, perfect. I'll add two explicit tests to cover the new behavior of future and past schedules.

tpl/Dashboard/resource_availability.tpl
```diff
@@ -61,6 +58,17 @@
...
+                            {* Determine if this is a future schedule in Unavailable section *}
+                            {assign var=isFutureScheduleUnavailable value=false}
+                            {assign var=dateFormat value='dashboard'}
+                            {if $section.dateField == 'ReservationEnds' && $s->HasAvailability()}
+                                {assign var=scheduleBegin value=$s->GetAvailabilityBegin()}
+                                {if $date == $scheduleBegin}
+                                    {assign var=isFutureScheduleUnavailable value=true}
+                                    {assign var=dateFormat value='schedule_daily'}
+                                {/if}
+                            {/if}
```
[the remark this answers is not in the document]
labmecanicatec: START_DATE is added to any item in the Unavailable section.

tpl/Dashboard/resource_availability.tpl
```diff
@@ -92,7 +100,7 @@
                                 {* Button *}
                                 <div class="reserveButton col-12 col-sm-3 d-grid">
                                     <button class="btn btn-sm btn-primary"
-                                        onclick="window.location.href='{$Path}{Pages::RESERVATION}?{QueryStringKeys::RESOURCE_ID}={$i->ResourceId()}{if $section.urlDate && $date}&{QueryStringKeys::START_DATE}={format_date date=$date timezone=$Timezone key=url_full}{/if}'">
+                                        onclick="window.location.href='{$Path}{Pages::RESERVATION}?{QueryStringKeys::RESOURCE_ID}={$i->ResourceId()}{if $isFutureScheduleUnavailable && $date}&{QueryStringKeys::START_DATE}={format_date date=$date timezone=$Timezone key=url}{/if}'">
                                         {translate key=Reserve}
```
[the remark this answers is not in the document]
labmecanicatec: START_DATE is added to any item in the Unavailable section.

tpl/Dashboard/resource_availability.tpl
```diff
@@ -61,6 +58,17 @@
                                 {assign var=date value=$i->ReservationEnds()}
                             {/if}
 
+                            {* Determine if this is a future schedule in Unavailable section *}
+                            {assign var=isFutureScheduleUnavailable value=false}
```
JohnVillalovos: Is this used?

tpl/Dashboard/resource_availability.tpl
```diff
@@ -61,6 +58,17 @@
...
+                            {assign var=dateFormat value='dashboard'}
+                            {if $section.dateField == 'ReservationEnds' && $s->HasAvailability()}
+                                {assign var=scheduleBegin value=$s->GetAvailabilityBegin()}
+                                {if $date == $scheduleBegin}
+                                    {assign var=isFutureScheduleUnavailable value=true}
```
JohnVillalovos: Is this used?

Presenters/Dashboard/ResourceAvailabilityControlPresenter.php
JohnVillalovos: Should this use `$schedules` from line 47? So there isn't two database calls.

Presenters/Dashboard/ResourceAvailabilityControlPresenter.php
```diff
@@ -50,6 +58,30 @@ public function PageLoad(UserSession $user)
...
+                    $end = $schedule->GetAvailabilityEnd();
+
+                    // If schedule is past, don't show this resource
+                    if ($now->GreaterThanOrEqual($end)) {
+                        continue;
+                    }
+
+                    // If schedule is future, add to unavailable with availability begin date
+                    if ($now->LessThan($begin)) {
+                        $futureReservation = new ReservationItemView(null, $begin, $begin, $resource->GetName(), $resource->GetId());
```
JohnVillalovos: Please use named arguments

tpl/Dashboard/resource_availability.tpl
labmecanicatec: No, actually $isFutureScheduleUnavailable is not used

tpl/Dashboard/resource_availability.tpl
labmecanicatec: No, actually $isFutureScheduleUnavailable is not used

Presenters/Dashboard/ResourceAvailabilityControlPresenter.php
labmecanicatec: Yes, it's better to reuse $schedules

Presenters/Dashboard/ResourceAvailabilityControlPresenter.php
labmecanicatec: Done!

JohnVillalovos:
# Codex Review of commit `f32659305`

## Findings

### P2 — The reported failure path lacks regression coverage

The new test only verifies presenter classification and the synthetic date in
`tests/Presenters/Dashboard/ResourceAvailabilityControlPresenterTest.php:87`.
The actual fix for issue #1512 depends on
`tpl/Dashboard/resource_availability.tpl:101` generating a valid `sd` query
parameter when the user clicks **Reserve**.

Add a template or entry-point regression test that renders or follows that URL
and confirms that the reservation page receives valid, populated dates. Without
that coverage, the link conditional or date format can regress and reproduce the
original crash while all added tests remain green.

## Recommendation

Do not merge yet. Add committed regression coverage for the generated
reservation link and reservation-page initialization. The implementation
otherwise looks sound.

JohnVillalovos:
# Claude code review: `f32659305` — fix(dashboard): handle schedule availability windows

## Findings

### 1. The presenter/template contract is implicit — `tpl/Dashboard/resource_availability.tpl:61-68`

The template recovers "is this a future-schedule placeholder?" by testing
`$date == $scheduleBegin`. The presenter knows this fact for certain and throws
it away; the template guesses it back with a loose object comparison.

It works today (the synthetic item's `EndDate` *is* the same `Date` instance
`GetAvailabilityBegin()` returns), and I could not construct a realistic false
positive — a real reservation would have to end at exactly the availability-begin
instant, which requires `begin == now` to the second. So it is **not a live
bug**, but it is an unenforced coupling that breaks silently the moment
availability semantics change.

The presenter should expose it: an `IsFutureAvailability()` on
`UnavailableDashboardItem`, or a separate bucket.

This is the main blocker.

### 2. Mislabeled named argument — `ResourceAvailabilityControlPresenter.php:76-82`

```php
new ReservationItemView(
    reservationId: null,
    startDate: $begin,
    ...
```

`reservationId` is the **6th** constructor parameter
(`Domain/ReservationItemView.php:440-461`); the first is `$referenceNumber`.

Harmless in practice — both parameters default to `null` and both end up `null`,
so behavior is identical and the tests can't catch it. But the argument claims to
mean something other than what it does, which defeats the point of naming it.
Since `null` is already the default, drop it entirely:

```php
new ReservationItemView(
    startDate: $begin,
    endDate: $begin,
    resourceName: $resource->GetName(),
    resourceId: $resource->GetId()
);
```

### 3. Redundant code — `ResourceAvailabilityControlPresenter.php:4,84`

`ReservationItemView::__construct` already ends with
`$this->UserPreferences = UserPreferences::Parse($preferences);`
(`Domain/ReservationItemView.php:536`), and `Parse(null)` returns
`new UserPreferences()`. Both the explicit assignment on line 84 and the
`require_once` on line 4 are no-ops and should be deleted.

### 4. Undoes a recent refactor without cause — `tpl/Dashboard/resource_availability.tpl:14-33`

Commit `82f14da1a` ("refactor(availability): simplify template using section
config loop") introduced the `urlDate` flag as an intent-revealing key. This
commit deletes it and replaces both call sites with
`$section.dateField == 'ReservationEnds'`. Behaviorally identical (both
unavailable sections use that field), but it swaps a named flag for a
stringly-typed check in two places, and it is unrelated to the bug being fixed.

### 5. Test gaps

- No test for `SetAvailableAllYear()` / `HasAvailability() === false` — the
  default production configuration, and the branch most at risk of regressing
  into "everything disappears from the dashboard."
- No boundary tests: `now == end` (hidden) and `now == begin` (active). Worth
  pinning, especially since `ScheduleAvailabilityRule`
  (`lib/Application/Reservation/Validation/ScheduleAvailabilityRule.php:43-44`)
  treats the end as *inclusive* (`LessThanOrEqual`) while this code treats it as
  exclusive (`GreaterThanOrEqual($end)` → hide).
- No test for a resource whose `ScheduleId` isn't in `GetAll()` (the `isset`
  fallback path).
- All the new template logic is untested —
  `FakeResourceAvailabilityControl::SetSchedules()` is a `// TODO: Implement`
  no-op.

### 6. The test churn is treating a symptom — `tests/fakes/FakeSchedule.php:13`

The root cause is that `FakeSchedule` hardcodes
`SetAvailability(2018-01-01, 2018-02-02)` in its constructor. Overriding it
inside this one test's `PopulateSchedules()` fixes this file but leaves the
landmine for every other suite that starts exercising this code path.
`SetAvailableAllYear()` as the fake's default would be the real fix (worth
checking the blast radius first).

Also, the new tests hardcode `2026-07-29`, which reads like "today when I wrote
this"; the existing test in the same file uses a neutral `2017-01-20`.

### 7. Scope — the link change is a separate concern

Repointing the resource name from `reservation.php` to `schedule.php` applies to
**all three sections**, not just future schedules, and additionally injects
`sd=<reservation end date>` for both unavailable sections — so clicking a
currently-reserved resource now jumps the schedule to a future date instead of
today. The commit body does describe this, but it is a UX decision bundled into
a bug fix and can't be reverted independently. Recommend splitting.

The mechanics of the new link are sound: `sid`/`rid`/`sd` all exist
(`lib/Server/QueryStringKeys.php:59,66,72`), `SchedulePageBuilder.php:365-368`
honors `rid` as a resource filter, `Pages::SCHEDULE` exists, both date-format
keys exist (`lang/en_us.php:22,27`), the raw `&` matches existing template
convention, `resourcePopup.js` defaults `preventClick: false` so the anchor still
navigates, and `Date::Compare` (`lib/Common/Date.php:270-276`) normalizes
timezones. The objection is purely about scope.

### 8. Worth verifying manually

For a future schedule the Reserve button pre-fills `sd = availability begin at
00:00` (the admin UI parses a date-only input —
`Presenters/Admin/ManageSchedulesPresenter.php:620`), which is likely outside the
schedule's daily open hours. Confirm the reservation page snaps to a valid slot
rather than erroring.

## Verdict: should this be merged upstream?

**Not yet.** The bug is real and the fix is in the right place.

**Must fix**

- Move future-detection into the presenter — the date-equality heuristic is the
  one real design problem (#1)
- Drop or correct the mislabeled `reservationId:` argument (#2)
- Remove the redundant `UserPreferences` assignment and `require_once` (#3)

**Should fix**

- A test for the no-availability-window (all-year) case (#5)
- Either restore `urlDate` or justify its removal (#4)

**Needs a maintainer decision**

- Whether hiding past-schedule resources outright is what #1512 asked for —
  unlike the future case, it gives the user no explanation
- Whether the resource-name link change belongs in this commit at all (#7)

JohnVillalovos (approved):
Thanks @labmecanicatec 

LGTM

PR #1624, opened 2026-08-02
feat(calendar): show a descriptive name in ICS subscription feeds

lucs7:
Add per-subscription calendar name (schedule name, resource name, resource group name, or the current user's own calendar) and pass it to CalendarExportDisplay::Render() so subscribed calendars ics files get a readable NAME/X-WR-CALNAME instead of a generic one in calendar clients.

### DIFF
--- a/Pages/Export/ICalendarSubscriptionPage.php
+++ b/Pages/Export/ICalendarSubscriptionPage.php
@@ -17,6 +17,8 @@ interface ICalendarSubscriptionPage
      */
     public function SetReservations($reservations);
 
+    public function SetCalendarName(string $calendarName): void;
+
     /**
      * @return string
      */
--- a/Pages/Export/SubscriptionPage.php
+++ b/Pages/Export/SubscriptionPage.php
@@ -24,6 +24,8 @@ abstract class SubscriptionPage extends Page implements ICalendarSubscriptionPag
     /** @var iCalendarReservationView[] */
     protected array $reservations = [];
 
+    protected ?string $calendarName = null;
+
     protected function __construct()
     {
         $authorization = new ReservationAuthorization(PluginManager::Instance()->LoadAuthorization());
@@ -42,6 +44,11 @@ abstract class SubscriptionPage extends Page implements ICalendarSubscriptionPag
     public function SetReservations($reservations): void
     {
         $this->reservations = $reservations;
+    }
+
+    public function SetCalendarName(string $calendarName): void
+    {
+        $this->calendarName = $calendarName;
     }
 
     public function GetSubscriptionKey()
--- a/lib/Application/Schedule/CalendarSubscriptionService.php
+++ b/lib/Application/Schedule/CalendarSubscriptionService.php
@@ -47,6 +47,15 @@ interface ICalendarSubscriptionService
      * @return int[]
      */
     public function GetResourcesInGroup($publicResourceGroupId);
+
+    public function GetResourceGroup(string $publicResourceGroupId): ?ResourceGroup;
+
+    public function ResolveCalendarName(
+        ?string $publicScheduleId,
+        ?string $publicResourceId,
+        ?string $publicResourceGroupId,
+        ?string $publicUserId = null
+    ): ?string;
 }
 
 class CalendarSubscriptionDetails
@@ -106,6 +115,7 @@ class CalendarSubscriptionDetails
 class CalendarSubscriptionService implements ICalendarSubscriptionService
 {
     private $cache = [];
+    private $groupCache = [];
 
     /**
      * @var IUserRepository
@@ -177,7 +187,7 @@ class CalendarSubscriptionService implements ICalendarSubscriptionService
     public function GetResourcesInGroup($publicResourceGroupId)
     {
         if (!array_key_exists($publicResourceGroupId, $this->cache)) {
-            $group = $this->resourceRepository->LoadResourceGroupByPublicId($publicResourceGroupId);
+            $group = $this->LoadResourceGroup($publicResourceGroupId);
 
             if ($group == null) {
                 return [];
@@ -188,6 +198,51 @@ class CalendarSubscriptionService implements ICalendarSubscriptionService
         }
 
         return $this->cache[$publicResourceGroupId];
+    }
+
+    public function GetResourceGroup(string $publicResourceGroupId): ?ResourceGroup
+    {
+        return $this->LoadResourceGroup($publicResourceGroupId);
+    }
+
+    public function ResolveCalendarName(
+        ?string $publicScheduleId,
+        ?string $publicResourceId,
+        ?string $publicResourceGroupId,
+        ?string $publicUserId = null
+    ): ?string {
+        $calendarName = null;
+
+        if (!empty($publicScheduleId)) {
+            $calendarName = $this->GetSchedule($publicScheduleId)->GetName();
+        }
+
+        if (!empty($publicResourceId)) {
+            $calendarName = $this->GetResource($publicResourceId)->GetName();
+        }
+
+        if (!empty($publicResourceGroupId)) {
+            $resourceGroup = $this->GetResourceGroup($publicResourceGroupId);
+            if ($resourceGroup !== null && $resourceGroup->name !== null && $resourceGroup->name !== '') {
+                $calendarName = $resourceGroup->name;
+            }
+        }
+
+        if (!empty($publicUserId)) {
+            $myCalendarName = Resources::GetInstance()->GetString('MyCalendar');
+            $calendarName = empty($calendarName) ? $myCalendarName : sprintf('%s - %s', $myCalendarName, $calendarName);
+        }
+
+        return $calendarName;
+    }
+
+    private function LoadResourceGroup(string $publicResourceGroupId): ?ResourceGroup
+    {
+        if (!array_key_exists($publicResourceGroupId, $this->groupCache)) {
+            $this->groupCache[$publicResourceGroupId] = $this->resourceRepository->LoadResourceGroupByPublicId($publicResourceGroupId);
+        }
+
+        return $this->groupCache[$publicResourceGroupId];
     }
 
     public function ForUser($userId, $resourceId = null, $scheduleId = null)
--- a/Pages/Export/CalendarSubscriptionPage.php
+++ b/Pages/Export/CalendarSubscriptionPage.php
@@ -21,6 +21,6 @@ class CalendarSubscriptionPage extends SubscriptionPage
         header('Content-Disposition: inline; filename=calendar.ics');
 
         $display = new CalendarExportDisplay();
-        echo $display->Render($this->reservations);
+        echo $display->Render($this->reservations, $this->calendarName);
     }
 }
--- a/Presenters/CalendarSubscriptionPresenter.php
+++ b/Presenters/CalendarSubscriptionPresenter.php
@@ -84,27 +84,34 @@ class CalendarSubscriptionPresenter
         $summaryFormat = Configuration::Instance()->GetKey(ConfigKeys::RESERVATION_LABELS_ICS_SUMMARY);
 
         $reservationUserLevel = ReservationUserLevel::OWNER;
+
         if (!empty($scheduleId)) {
-            $schedule = $this->subscriptionService->GetSchedule($scheduleId);
-            $sid = $schedule->GetId();
+            $sid = $this->subscriptionService->GetSchedule($scheduleId)->GetId();
         }
         if (!empty($resourceId)) {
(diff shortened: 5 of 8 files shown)


lib/Application/Schedule/CalendarSubscriptionService.php
```diff
@@ -190,6 +196,13 @@ public function GetResourcesInGroup($publicResourceGroupId)
         return $this->cache[$publicResourceGroupId];
     }
 
+    public function GetResourceGroupName($publicResourceGroupId)
+    {
+        $group = $this->resourceRepository->LoadResourceGroupByPublicId($publicResourceGroupId);
+
+        return $group === null ? null : $group->name;
```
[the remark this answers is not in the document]
lucs7: both GetResourcesInGroup() and GetResourceGroupName() now go through a shared LoadResourceGroup() helper backed by a dedicated $groupCache

lucs7:
the resource resource-group filter was present in the code and is not touched by the PR. I created a separate bug report for later fixing

lib/Application/Schedule/CalendarSubscriptionService.php
```diff
@@ -47,6 +47,12 @@ public function ForSchedule($scheduleId);
      * @return int[]
      */
     public function GetResourcesInGroup($publicResourceGroupId);
+
+    /**
+     * @param string $publicResourceGroupId
+     * @return null|string
+     */
+    public function GetResourceGroupName($publicResourceGroupId);
```
JohnVillalovos: Can this use 'type-hints' instead of a docblock? Trying to use type-hints where possible, though not always possible.

Pages/Export/ICalendarSubscriptionPage.php
```diff
@@ -17,6 +17,11 @@ public function GetUserId();
      */
     public function SetReservations($reservations);
 
+    /**
+     * @param string $calendarName
+     */
+    public function SetCalendarName($calendarName);
```
JohnVillalovos: Can this use 'type-hints' instead of a docblock? Trying to use type-hints where possible, though not always possible.

lib/Application/Schedule/CalendarSubscriptionService.php
```diff
@@ -190,6 +197,22 @@ public function GetResourcesInGroup($publicResourceGroupId)
         return $this->cache[$publicResourceGroupId];
     }
 
+    public function GetResourceGroupName($publicResourceGroupId)
```
JohnVillalovos: Please add type-hints

Pages/Export/SubscriptionPage.php
```diff
@@ -44,6 +46,11 @@ public function SetReservations($reservations): void
         $this->reservations = $reservations;
     }
 
+    public function SetCalendarName($calendarName): void
```
JohnVillalovos: Please add type-hints

JohnVillalovos:
# Claude Code review: `f0fa2a4a6` — feat(calendar): show a descriptive name in ICS subscription feeds

## What's good

- Correct layering: presenter decides the name, page holds it, display renders
  it. No new coupling introduced.
- `LoadResourceGroup()` + `$groupCache` avoids a second
  `LoadResourceGroupByPublicId` round trip, and
  `testGetResourceGroupNameReusesGroupLoadedByGetResourcesInGroup` actually
  asserts the repository is hit exactly once. Good test.
- The group-name null/empty guard correctly avoids clobbering an
  already-derived schedule name, with a dedicated test
  (`testGetsResourceGroupReservationsKeepsScheduleCalendarNameWhenGroupNameIsNull`).
- No new information disclosure. The feed already exposes reservation detail
  behind the subscription key, and the personal feed uses a generic label rather
  than the user's name.
- sabre/vobject escapes and folds the property value, so an admin-authored
  resource name containing commas, semicolons or newlines can't corrupt the ICS.

## Findings

### 1. Precedence makes personal filtered feeds indistinguishable

`Presenters/CalendarSubscriptionPresenter.php:103-109`

The `!empty($userId)` block runs *after* the schedule and resource blocks, so it
overwrites their names. That matters because
`Presenters/Calendar/PersonalCalendarPresenter.php:57` calls
`ForUser($userSession->UserId, $resourceId, $scheduleId)`, and
`CalendarSubscriptionUrl::__construct()` puts **all three** (`uid`, `sid`, `rid`)
into the generated URL.

Result: a user who subscribes to "my calendar filtered to Room A" and "my
calendar filtered to Room B" gets two feeds both named `My Calendar` — which is
the exact problem this commit sets out to solve.

A composed name (`My Calendar – Room A`) or letting the resource/schedule name
win when present would fix it. Not a merge blocker, but it's the one case where
the feature under-delivers, and it's untested — no test covers `uid` + `rid`
together.

### 2. No test asserts the name reaches the ICS output

Nothing under `tests/` asserts `X-WR-CALNAME` for an export or subscription;
the only occurrence is `tests/Presenters/Admin/ImportICalPresenterTest.php:108`,
which is a parsing fixture. The wiring is tested at the presenter/page boundary,
but the payload never is.

`tests/Presenters/CalendarExportPresenterTest.php` already instantiates
`CalendarExportDisplay` and inspects serialized output in five places, so a
one-line assertion there would close the gap cheaply.

### 3. Minor: naming/consistency in the service

`lib/Application/Schedule/CalendarSubscriptionService.php:200`

Sibling methods return domain objects (`GetResource`, `GetSchedule`, `GetUser`).
A `GetResourceGroup($publicId): ?ResourceGroup` would fit the existing shape
better and let callers take what they need, rather than adding a name-specific
accessor to the public interface. Cosmetic; wouldn't block on it.

### 4. Docblocks used where native type declarations would work

The commit adds `@param` / `@return` docblocks on brand-new declarations instead
of native types. CLAUDE.md asks for type declarations on new functions and
methods, and the author already used a native type for the backing property
(`protected ?string $calendarName = null`, `Pages/Export/SubscriptionPage.php:27`),
so the two styles are inconsistent within the same commit.

**Native types cannot be adopted one line at a time here.** PHP rejects partial
adoption in both directions — verified on PHP 8.4.17:

- Interface param untyped, implementation narrowing to `?string` →
  `Fatal error: Declaration of CA::f(?string $x): void must be compatible with IA::f($x)`
- Interface declaring `: ?string`, implementation omitting it →
  `Fatal error: Declaration of CB::g($x) must be compatible with IB::g($x): ?string`

So the interface and every implementer must be typed in the same change. The
table below therefore lists the interface and its implementation as separate
rows carrying the *same* signature — PHP has no type inheritance, so each
declaration stands on its own.

**The two halves are not equally optional.** The return type is mandatory on the
implementation (second bullet above). The parameter type is not — an untyped
param is `mixed`, which is wider than `string`, so contravariance lets the class
load. But PHP checks arguments against the signature of the function actually
invoked, not the interface it satisfies, so a half-typed pair silently stops
enforcing the contract, *including* for calls made through the interface type:

```php
interface I3 { public function f(string $x): ?string; }
class C3 implements I3 { public function f($x): ?string { var_dump($x); return null; } }

$c->f(123);                              // int(123)  - no error
(function (I3 $i) { $i->f(456); })($c);  // int(456)  - no error, called via the interface
```

Typed on both sides, the same call is rejected with
`TypeError: C4::f(): Argument #1 ($x) must be of type string, array given`.
That makes the half-typed form the quietly-broken case rather than a build
failure, which is the reason to insist on both:

| Interface | Implementation | Result |
|---|---|---|
| `f(string $x): ?string` | `f(string $x): ?string` | Correct — enforced at runtime |
| `f(string $x): ?string` | `f($x): ?string` | Loads, but param contract is unenforced |
| `f(string $x): ?string` | `f(string $x)` | Fatal error at declaration |

For this commit that is six declarations across four files:

| Site | Current | Typed equivalent |
|---|---|---|
| `Pages/Export/ICalendarSubscriptionPage.php:20-23` | `@param string $calendarName` + untyped | `SetCalendarName(?string $calendarName): void` |
| `Pages/Export/SubscriptionPage.php:49` | `SetCalendarName($calendarName): void` | `SetCalendarName(?string $calendarName): void` |
| `tests/Presenters/CalendarSubscriptionPresenterTest.php:302` | `SetCalendarName($calendarName)` (the fake) | `SetCalendarName(?string $calendarName): void` |
| `lib/Application/Schedule/CalendarSubscriptionService.php:51-55` | `@param string` / `@return null\|string` + untyped | `GetResourceGroupName(string $publicResourceGroupId): ?string` |
| `lib/Application/Schedule/CalendarSubscriptionService.php:200` | untyped | `GetResourceGroupName(string $publicResourceGroupId): ?string` |
| `lib/Application/Schedule/CalendarSubscriptionService.php:207` | untyped (private) | `LoadResourceGroup(string $publicResourceGroupId): ?ResourceGroup` |

**This stays inside the commit's own blast radius.** Every one of those six is a
`+` line introduced by this commit, in a file it already modifies — no
pre-existing declaration has to change. Confirmed by `git show HEAD` and by
grepping the tree: the only three `SetCalendarName` declarations that exist are
the interface, `SubscriptionPage`, and the test fake. `AtomSubscriptionPage` and
`CalendarSubscriptionPage` inherit the concrete method without redeclaring it,
and `CalendarSubscriptionValidatorTest` uses `createMock`, which generates
type-compatible implementations at runtime.

`string` rather than `?string` is correct for `$publicResourceGroupId`: the sole
caller guards with `!empty($resourceGroupId)` before calling
(`Presenters/CalendarSubscriptionPresenter.php:111-113`), and querystring values
are strings. `?ResourceGroup` is safe for the private helper because
`LoadResourceGroupByPublicId()` returns a `ResourceGroup` or null
(`Domain/Access/ResourceRepository.php:590-603`).

**Where the docblocks should stay.** The neighbouring untyped members —
`SetReservations()`, `GetResourcesInGroup()`, `GetResource()`, `GetSchedule()`,
`GetUser()` — are pre-existing declarations with implementers and callers
outside this commit's diff (e.g. `CalendarExportPage::SetReservations()`).
Typing those is a separate refactor and correctly out of scope here; the
author's "match local style" reasoning holds for them, just not for the six new
declarations above.

Note this interacts with finding #5: typing `ICalendarSubscriptionPage` widens
the ABI break for out-of-tree implementers beyond the method addition alone.

### 5. Note, not a defect: interface backwards compatibility

Adding `SetCalendarName()` to `ICalendarSubscriptionPage` and
`GetResourceGroupName()` to `ICalendarSubscriptionService` breaks any
out-of-tree implementer. Nothing in-tree breaks, and this is consistent with how
the project has evolved these interfaces. Worth a mention only if the
maintainers care about third-party page/service implementations.

### 6. Unrelated to the commit, but in the working tree

`tpl/Export/ical.tpl` is untracked and dead after the sabre/vobject migration.
Make sure it doesn't get swept into a commit.

## Verdict

**Yes, merge it** — but raise finding #1 on the PR first.

The change is well-scoped, unit-tested, and passes every CI gate locally. The
precedence question deserves a maintainer decision before merge (a composed name
is a small follow-up either way), and adding one `X-WR-CALNAME` assertion to
`CalendarExportPresenterTest` would close the only real coverage gap.

JohnVillalovos:
# Codex Review of `f0fa2a4`

One non-blocking maintainability finding.

## Finding

### Low: Use native types for the newly introduced APIs

The new methods rely on PHPDoc where native PHP type declarations can express the contracts directly:

- `ICalendarSubscriptionPage::SetCalendarName()` documents a `string` parameter but has no parameter or return type.
- `CalendarSubscriptionService::GetResourceGroupName()` documents `null|string` but has no native parameter or return type.
- The new private `LoadResourceGroup()` helper is also untyped.

With PHP 8.2 as the minimum supported version and the repository guidance to use type hints where possible, these should be declared as:

```php
public function SetCalendarName(string $calendarName): void;
public function GetResourceGroupName(string $publicResourceGroupId): ?string;
private function LoadResourceGroup(string $publicResourceGroupId): ?ResourceGroup;
```

The concrete `SubscriptionPage` implementation and the test fake should use the same `SetCalendarName()` signature. Native declarations make these new contracts enforceable and avoid PHPDoc drifting from the implementation.

## Recommendation

Merge upstream after addressing the native-type cleanup above. It is small and does not require a design change.

## Review summary

- Correctly selects schedule, resource, user, and resource-group calendar names.
- Resource-group lookup caching avoids duplicate repository calls.
- Calendar names reach both `NAME` and `X-WR-CALNAME` in the serialized ICS feed.
- The commit message follows Conventional Commits, has a 65-character header, and includes the required `Assisted-by` trailer.

## Additional non-blocking note

There is no automated page-level test covering the final forwarding call in `Pages/Export/CalendarSubscriptionPage.php`. The implementation is straightforward, and the direct serialization smoke test confirmed the expected output, so this does not block merging.

JohnVillalovos (approved):
LGTM

Thanks @lucs7

PR #1629, opened 2026-08-03
fix(schedule): don't require resource permissions for anonymous label formatting

### LINKED ISSUE #1616: [Bug] ICS calendar subscription: event SUMMARY (title) is empty in 5.2.0
### Acknowledgements

### Describe the bug

[This bug/issue and especially the root cause was examined with the help of Claude Opus 4.8]

Since upgrading to 5.2.0, subscribed ICS calendar feeds (webcal / CalendarSubscription) emit an empty SUMMARY for every VEVENT. All other fields (DESCRIPTION, ORGANIZER, LOCATION) are populated correctly — only the event title is missing, so calendar clients show untitled events.

This worked in 5.1.0. The reservations do have titles (they display normally in the web UI), and it is not a privacy or label-configuration problem (see root cause below).

**Environment**

- LibreBooking 5.2.0 (official librebooking/librebooking:5.2.0 Docker image)
- Feed accessed as a normal calendar subscription, i.e. fetched by the client without an authenticated session
- Privacy settings permissive: view.reservations = true, hide.reservation.details = false, hide.user.details = false
- reservation.labels.ics.summary set to a non-empty template (default {title} also reproduces)

### To Reproduce

1. Create a reservation with a title on any resource.
2. Subscribe to a schedule/resource/personal ICS feed and fetch it unauthenticated, e.g.:
   `curl -sL 'https://<host>/<ics-subscription-url>' | grep -E 'BEGIN:VEVENT|SUMMARY|DESCRIPTION'`
3. Observe that SUMMARY: is present but empty, while DESCRIPTION: and ORGANIZER/LOCATION are populated.

### Expected behavior

`SUMMARY` contains the reservation title (as in 5.1.0).

**Actual**

`SUMMARY` is an empty string for every event.

### Screenshots

_No response_

## Root cause

In `lib/Application/Schedule/iCalendarReservationView.php`, the way `Summary` is built changed between releases.

**5.1.0** — title read directly, gated only by the ICS view's own privacy check:
```php
$this->Summary = $canViewDetails ? $res->Title : $privateNotice;
```

**5.2.0** — title now routed through `SlotLabelFactory::Format()`:
```php
$this->Summary = $canViewDetails ? self::toRfc5545Text($factory->Format($res, $summaryFormat)) : $privateNotice;
```

`SlotLabelFactory::Format()` (in `lib/Application/Schedule/SlotLabelFactory.php`) is the interactive-schedule label renderer. It contains a per-user gate that returns an empty string when the **requesting** user has no permission to the reservation's resource and is not its owner/invitee/participant:
```php
if (!in_array($reservation->ResourceId, $this->UserResourcePermissions($this->user->UserId))
    && !$reservation->IsUserOwner($this->user->UserId)
    && !$reservation->IsUserInvited($this->user->UserId)
    && !$reservation->IsUserParticipating($this->user->UserId)) {
    return '';
}
```

`CalendarSubscriptionPresenter` passes the current server session into the view, and for a webcal subscription that session is **anonymous** — it holds no resource permissions and owns nothing — so `Format()` returns `''` and the summary is blank.

The other iCal fields (`DESCRIPTION` from `$res->Description`, `ORGANIZER`, `LOCATION`) are assigned directly and bypass this factory, which is exactly why every field except the title still renders. The `iCalendarReservationView` already performs its own, ICS-appropriate visibility check via `$canViewDetails` (the `IPrivacyFilter`), so re-gating the summary through the schedule label factory double-applies a stricter, login-dependent rule that doesn't fit an unauthenticated export.

## Suggested fix

The ICS summary should not be subject to `SlotLabelFactory`'s per-user resource-permission gate. Simplest option — restore the direct title read while keeping 5.2.0's RFC-5545 escaping:

```diff
- $this->Summary = $canViewDetails ? self::toRfc5545Text($factory->Format($res, $summaryFormat)) : $privateNotice;
+ $this->Summary = $canViewDetails ? self::toRfc5545Text($res->Title ?? '') : $privateNotice;
```

If the configurable `ics.summary` template should still be honored, the token substitution needs to run without the login-based early-returns in `SlotLabelFactory::Format()` (e.g. a "trusted"/export mode, or a dedicated formatter), gated solely by the `$canViewDetails` already computed in `iCalendarReservationView`.

## Workaround

Overlaying a patched `iCalendarReservationView.php` with the one-line change above restores correct titles in the feed.

### LibreBooking version

5.2.0

lucs7:
SlotLabelFactory::Format() unconditionally required the requesting user to hold resource permissions or be owner/invitee/participant before rendering a reservation label. Anonymous sessions (UserId 0) can never satisfy that check, so every anonymous label render returned an empty string.

This broke unauthenticated ICS calendar subscriptions (SUMMARY was always blank, while DESCRIPTION/ORGANIZER/LOCATION rendered fine since they bypass this factory), even though privacy.view.reservations already permits anonymous viewing and iCalendarReservationView had already computed that via its own privacy filter. The same gate is also reached by the public schedule/calendar views and the Atom feed, which deliberately pass an anonymous session when guest viewing is enabled.

Skip the resource-permission check for non-logged-in users, since anonymous access is already gated earlier in the same method by privacy.view.reservations.

Closes: #1616

### DIFF
--- a/lib/Application/Schedule/SlotLabelFactory.php
+++ b/lib/Application/Schedule/SlotLabelFactory.php
@@ -72,7 +72,13 @@ class SlotLabelFactory
             return '';
         }
 
-        if (!in_array($reservation->ResourceId, $this->UserResourcePermissions($this->user->UserId)) && !$reservation->IsUserOwner($this->user->UserId) && !$reservation->IsUserInvited($this->user->UserId) && !$reservation->IsUserParticipating($this->user->UserId)) {
+        if (
+            $this->user->IsLoggedIn() &&
+            !in_array($reservation->ResourceId, $this->UserResourcePermissions($this->user->UserId)) &&
+            !$reservation->IsUserOwner($this->user->UserId) &&
+            !$reservation->IsUserInvited($this->user->UserId) &&
+            !$reservation->IsUserParticipating($this->user->UserId)
+        ) {
             return '';
         }
 
--- a/tests/Application/Schedule/SlotLabelFactoryTest.php
+++ b/tests/Application/Schedule/SlotLabelFactoryTest.php
@@ -168,6 +168,19 @@ class SlotLabelFactoryTest extends TestBase
         $this->assertEquals('', $label);
     }
 
+    public function testAnonymousUserSeesLabelWhenPublicViewingIsAllowed()
+    {
+        $this->SetConfig('{name} - {title}');
+        $this->fakeConfig->SetKey(ConfigKeys::PRIVACY_VIEW_RESERVATIONS, 'true');
+
+        $anonymousUser = new NullUserSession();
+        $factory = new SlotLabelFactory($anonymousUser);
+
+        $label = $factory->Format($this->reservation);
+
+        $this->assertEquals('first last - some title', $label);
+    }
+
     private function SetConfig($value)
     {
         $this->fakeConfig->SetKey(ConfigKeys::SCHEDULE_RESERVATION_LABEL, $value);


JohnVillalovos (approved):
LGTM

Thanks @lucs7

PR #1631, opened 2026-08-04
chore: update select2 to 4.1.0

labmecanicatec:
Update Select2 assets to the latest stable version and keep
 the dependency aligned with current releases.


JohnVillalovos (approved):
I compared files to the upstream and everything matches.

LGTM

Thanks @labmecanicatec

Issue #1600, opened 2026-07-24
[Bug] Icons and schedule display problem on mobile browsers


belcirelk:
### Acknowledgements

- [x] This is a bug and not a question. Questions should be asked in the [Discussions area](https://github.com/LibreBooking/librebooking/discussions).
- [x] This project is maintained 100% by volunteers. I understand that there are no guarantees on when (or if) this will be addressed, and that the most effective way to see this resolved is for me to submit a Pull Request.

### Describe the bug

When using a mobile browser, the icons in schedule are wrong and 2 are missing on smaller screens. On a tablet, left screenshot, the standard schedule is displayed as the compact schedule. The second icon works as expected while the third gives the compact view. The compact view icon is working. 

On a phone (right), only 2 icons are displayed. Normal and long. The normal gives the compact view. 

https://github.com/user-attachments/assets/11bdb138-fe89-4b1c-b399-a4c82b545040

### To Reproduce

Go to a schedule on a mobile device, with a mobile browser. 

### Expected behavior

On tablets with mobile browser, 4 icons should be displayed with the same behavior as desktop. On phone, the icon should be the compact icon. Alternatively, both could have the same phone 2 icons display using the mobile mode of the browser (users have the option to use desktop mode).

### LibreBooking version

5.2 and +

JohnVillalovos:
@belcirelk Would be willing to re-test with the current `develop` branch and see what issues, if any, remain? Thanks.

belcirelk:
Fixed by #1642 and #1635

PR #1635, opened 2026-08-07
fix(schedule): correct schedule style icons on mobile and table. Fix …

asiimhusain:
What does this PR do? This PR fixes a bug where the schedule view icons in the toolbar behaved incorrectly or disappeared on mobile devices and tablets:

On Tablets: The Wide and Condensed Week schedule icons were completely missing because they were hidden by the CSS helper classes (d-none d-md-inline-flex) for viewports smaller than 768px (which includes portrait tablets). We changed these to d-none d-sm-inline-flex so that all 4 icons are visible and usable on tablets.
On Mobile Phones: Clicking the "Standard" layout icon (the table icon) correctly loads the mobile compact template layout, but the icon itself looked like the desktop standard grid. We made the icon and tooltip dynamically switch to display the Condensed/Compact icon on mobile phones to align the visual design with the actual view that gets rendered.
How was it tested?

Verified that on desktops and tablets, all 4 icons display and reload the schedule style correctly.
Verified that on mobile phones, the 2 icons display correctly with the first one using the compact layout icon.

### DIFF
--- a/tpl/Schedule/schedule.tpl
+++ b/tpl/Schedule/schedule.tpl
@@ -95,9 +95,9 @@
                                 <a href="#"
                                     class="schedule-style me-2 d-inline-flex align-items-center{if $ScheduleStyle == ScheduleStyle::Standard->value} active{/if}"
                                     id="schedule_standard" schedule-display="{ScheduleStyle::Standard->value}"
-                                    title="{translate key='StandardScheduleDisplay'}">
-                                    <img class="schedule_icon shadow-sm" src="img/table.png"
-                                        alt="{translate key='StandardScheduleDisplay'}" />
+                                    title="{if $IsMobile && !$IsTablet}{translate key='CondensedWeekScheduleDisplay'}{else}{translate key='StandardScheduleDisplay'}{/if}">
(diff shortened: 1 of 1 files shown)


belcirelk:
Related to bug #1600

asiimhusain:
> Related to bug #1600

yes it is.

belcirelk:
I have tested, it's definitely an improvement. 

For phone on vertical, it works perfectly. Tablet desktop view too. 


For me the view that doesn't work is on tablet with mobile browser (large enough for 4 buttons). But I usually use my browser in desktop mode on mobile. If you turn your phone, you get the same if you want to test.

labmecanicatec (commented):
For phones in landscape mode, the Condensed Week Schedule view is repeated in the first and fourth icons.

asiimhusain:
Hey,

Please check it now. The issue has been resolved.

If you notice anything else or the issue occurs again, just leave a comment here. I’m actively available to look into it and resolve it.

JohnVillalovos (approved):
Thanks @asiimhusain 

LGTM

PR #1643, opened 2026-08-09
fix(schedule): remove empty query strings

lucs7:
This pull request updates the `CalendarSubscriptionUrl` class to only include query string parameters for IDs that are actually provided, and adds tests to verify this behavior.

**Conditional query string parameter inclusion:**

* Updated the `__construct` method in `CalendarSubscriptionUrl` to only add `userPublicId`, `schedulePublicId`, and `resourcePublicId` to the query string if they are not empty.

**Test coverage improvements:**

* Added `CalendarSubscriptionUrlTest` to verify that only provided IDs are included in the generated URL, and that all provided IDs appear as expected.


JohnVillalovos (approved):
LGTM

Thanks @lucs7

PR #1644, opened 2026-08-09
fix(calendar): disable subscription link during loading

lucs7:
This pull request improves the user experience when subscribing to the calendar by disabling the "Subscribe to Calendar" link while a new subscription request is being processed. This prevents users from making duplicate requests or interacting with the link during loading.

### DIFF
--- a/Web/scripts/calendar.js
+++ b/Web/scripts/calendar.js
@@ -455,8 +455,20 @@ function Calendar(opts) {
     var url = _options.getSubscriptionUrl + '&rid=' + rid + '&sid=' + sid + '&gid=' + gid;
     ajaxGet(
       url,
-      function () {},
+      function () {
+        // Disable the link while it may still reflect the previous filter selection.
+        // The href and inline onclick are dropped as well: pointer-events: none only
+        // blocks the mouse, so a focused link could otherwise still be activated with
+        // the keyboard and copy the stale subscription URL to the clipboard.
+        $('#subscribeToCalendar')
+          .addClass('disabled')
+          .attr('aria-disabled', 'true')
+          .removeAttr('onclick')
+          .removeAttr('href');
+      },
       function (response) {
+        // Re-enabling happens by replacement: the link lives inside #calendarSubscription
+        // and is rendered fresh, enabled, and carrying the current filter's URL.
         $('#calendarSubscription').html(response);
       }
     );
--- a/Web/css/librebooking.css
+++ b/Web/css/librebooking.css
@@ -149,6 +149,11 @@ fieldset:disabled .btn {
   border-color: var(--primary-disabled);
 }
 
+a.disabled {
+  pointer-events: none;
+  opacity: 0.5;
+}
+
 .link-primary {
   color: var(--primary) !important;
   text-decoration: none;


Web/scripts/calendar.js
```diff
@@ -458,9 +458,13 @@ function Calendar(opts) {
     var url = _options.getSubscriptionUrl + '&rid=' + rid + '&sid=' + sid + '&gid=' + gid;
     ajaxGet(
       url,
-      function () {},
+      function () {
+        // Disable the link while it may still reflect the previous filter selection.
```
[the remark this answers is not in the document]
lucs7: actually if no done is achieved enabling the link would be wrong as it doesnt have the actual current rid/sid, so I would keep this

lucs7:
anything I can do here?

JohnVillalovos (commented):
Claude authored: Reviewed at `d74c092d7`. The race is real and the fix targets it correctly — `ajaxGet` maps `onBefore` to `beforeSend` and `onAfter` to `done` (`Web/scripts/ajax-helpers.js:22`), so the disabled window is exactly the request lifetime, and the stale-URL-to-clipboard window is genuinely closed for mouse users.

Two minor inline notes below; neither blocks merge. Copilot's keyboard/`aria-disabled` point on line 460 does still look worth addressing — see my note on line 464 for a gotcha about where that fix has to live.

The one thing I'd ask for before merge is a commit message body. The header is compliant, but there's no body and the *why* isn't obvious from the subject alone:

```
fix(calendar): disable subscription link during loading

Changing the calendar filter refreshes the subscription link over AJAX.
Until the response arrives, the link still carries the previous filter's
URL, so clicking it copies a stale subscription URL to the clipboard.
Disable the link for the duration of the request.
```

Plus a `Closes: #<issue>` footer if there's a tracking issue.

Web/scripts/calendar.js
```diff
@@ -455,9 +455,13 @@ function Calendar(opts) {
...
+        $('#subscribeToCalendar').addClass('disabled');
+      },
       function (response) {
         $('#calendarSubscription').html(response);
+        $('#subscribeToCalendar').removeClass('disabled');
```
JohnVillalovos: Claude authored: Minor: this `removeClass` is a no-op.

`#subscribeToCalendar` lives *inside* `#calendarSubscription` in both subscription templates, so `.html(response)` on the previous line has already destroyed the anchor that received the class. The replacement is rendered fresh from the template with `class="link-primary"` and never had `.disabled` on it. Re-enabling works by element replacement, not by this call.

It's harmless, and worth keeping as defense-in-depth if the markup ever changes — but as written the ordering reads as though the class survives the swap. A short comment, or dropping the line, would save a future reader the trace. Note that the same applies to whatever you add for Copilot's keyboard/`aria-disabled` point above: attributes set in the `beforeSend` callback won't survive the re-render either, so the fix for that needs to live in the template or be re-applied after `.html()`.

Web/css/schedule.css
```diff
@@ -480,3 +480,8 @@ img.schedule_icon {
 #reservations {
   position: relative;
 }
+
+#subscribeToCalendar.disabled {
```
JohnVillalovos: Claude authored: Style nit, non-blocking: this is ID-scoped in a page-specific stylesheet.

`Web/css/librebooking.css` already houses the project's disabled-state styling (`.btn.disabled`, `.disabled > .page-link`), so a generic `a.disabled` rule there would be reusable and more consistent with where this kind of thing lives. Bootstrap 5 deliberately does not style bare `a.disabled` — it scopes `.disabled` to `.btn`/`.nav-link`/`.page-link`/`.dropdown-item`/`.list-group-item` — so a shared rule would fill a real gap rather than duplicate one.

JohnVillalovos:
> anything I can do here?

@lucs7 If you can look at the comments and see if there is anything there, it would be appreciated. I'm not very good with the front-end stuff...

lucs7:
thank especially with the css placement I was wondering where it fit.

labmecanicatec (commented):
I think it would be better to render the link as a button directly in `tpl/Calendar/calendar.subscription.tpl`:

```html
<div id="calendarSubscription" class="calendar-subscription text-end">
    {if $IsSubscriptionAllowed && $IsSubscriptionEnabled}
        <button
            id="subscribeToCalendar"
            type="button"
            class="btn btn-sm btn-link link-primary"
            title="{translate key=SubscribeToCalendar}"
            onclick="copyUrlToClipboard('{$SubscriptionUrl|escape:'javascript'|escape:'html'}');"
        >
            <i class="bi bi-calendar-heart me-1"></i>{translate key=SubscribeToCalendar}
        </button>
    {/if}
</div>
```

Then the script could simply disable the button with:

```js
$('#subscribeToCalendar').prop('disabled', true);
```

This would avoid the need to add custom CSS styles. Also, since this is a button that performs an action rather than a link that navigates somewhere, I think using a `<button>` is more appropriate semantically.

On the other hand, I think the subscription links should use `title="{translate key=SubscribeToCalendar}"` instead.

labmecanicatec:
I think it would be better to render the link as a button directly in `tpl/Calendar/calendar.subscription.tpl`:

```html
<div id="calendarSubscription" class="calendar-subscription text-end">
    {if $IsSubscriptionAllowed && $IsSubscriptionEnabled}
        <button
            id="subscribeToCalendar"
            type="button"
            class="btn btn-sm btn-link link-primary"
            title="{translate key=SubscribeToCalendar}"
            onclick="copyUrlToClipboard('{$SubscriptionUrl|escape:'javascript'|escape:'html'}');"
        >
            <i class="bi bi-calendar-heart me-1"></i>{translate key=SubscribeToCalendar}
        </button>
    {/if}
</div>
```

Then the script could simply disable the button with:

```js
$('#subscribeToCalendar').prop('disabled', true);
```

This would avoid the need to add custom CSS styles. Also, since this is a button that performs an action rather than a link that navigates somewhere, I think using a `<button>` is more appropriate semantically.

On the other hand, I think the subscription links should use `title="{translate key=SubscribeToCalendar}"` instead.

JohnVillalovos:
@labmecanicatec Would you be okay with deferring your suggestion to a follow-up PR?

NOTE: I asked in Discord and @labmecanicatec was okay with deferring.

PR #1646, opened 2026-08-10
feat(ics): fix ICS subscription key generation if empty

lucs7:
This pull request improves the reliability and robustness of ICS (iCalendar) subscription key management in the application's configuration system. It ensures that when ICS subscriptions are enabled via the admin interface, a subscription key is automatically generated if one does not already exist, preventing broken subscription links. The changes also refactor and centralize the logic for generating the subscription key, and add comprehensive tests to verify correct behavior in various scenarios.

Key changes include:

### ICS Subscription Key Generation and Management

* Added a new private method `EnableIcsSubscriptionIfNeeded` in `ManageConfigurationPresenter.php` to automatically generate and inject an ICS subscription key into the settings array when ICS is enabled and no key is set, ensuring the key is always present when needed. [[1]](diffhunk://#diff-da3b7736ce68acb8149905fa05bb1bdd951a62bb433abdb47aaf77a0f0339debR227-R230) [[2]](diffhunk://#diff-da3b7736ce68acb8149905fa05bb1bdd951a62bb433abdb47aaf77a0f0339debR249-R280)
* Refactored `Configuration::EnableSubscription` and `ConfigurationFile::EnableSubscription` to accept an optional config file path, and to use a new static method `GenerateSubscriptionKeyIfNeeded` for consistent key generation logic. The key is now generated only if ICS is enabled and no key is set, and the config file is updated in-place using a regex replacement. [[1]](diffhunk://#diff-596cae96f35bfd2bdcdf597f59ce8761b2c204cd9e39bb02c74da5df52f30a55R265-R270) [[2]](diffhunk://#diff-596cae96f35bfd2bdcdf597f59ce8761b2c204cd9e39bb02c74da5df52f30a55R683-R736)
* Added the static method `GenerateSubscriptionKeyIfNeeded` to centralize the key generation logic and ensure consistent behavior across the codebase.

### Testing Improvements

* Added a new test class `EnableSubscriptionTest` to verify that the ICS subscription key is generated only when ICS is enabled and the key is empty, and that the config file is not modified in other cases (ICS disabled or key already set).
* Enhanced `ManageConfigurationPresenterTest` with tests to verify that the presenter generates a subscription key only when appropriate, and does not overwrite an existing key or generate a key when ICS is disabled.
* Updated `FakeConfig` to match the new method signature for `EnableSubscription`. [[1]](diffhunk://#diff-222b761c1c97f2e74e1e0898468e7b43e1aad1a931869d1fcdf69a79f0364253L42-R42) [[2]](diffhunk://#diff-222b761c1c97f2e74e1e0898468e7b43e1aad1a931869d1fcdf69a79f0364253L151-R151)

### DIFF
--- a/lib/Config/Configuration.php
+++ b/lib/Config/Configuration.php
@@ -63,10 +63,11 @@ interface IConfigurationFile
     public function GetAdminEmail();
 
     /**
-     * Enables ics subscriptionwith a random key if no key is set.
+     * Enables ics subscription with a random key if no key is set.
+     * @param string|null $configFilePath Path to rewrite; defaults to the live config file.
      * @return void
      */
-    public function EnableSubscription();
+    public function EnableSubscription(?string $configFilePath = null);
 }
 
 /**
@@ -262,11 +263,12 @@ class Configuration implements IConfiguration
 
     /**
      * Enables ICS subscription with a random key if no key is set.
+     * @param string|null $configFilePath Path to rewrite; defaults to the live config file.
      * @return void
      */
-    public function EnableSubscription()
+    public function EnableSubscription(?string $configFilePath = null)
     {
-        $this->File(self::DEFAULT_CONFIG_ID)->EnableSubscription();
+        $this->File(self::DEFAULT_CONFIG_ID)->EnableSubscription($configFilePath);
     }
 
     /**
@@ -679,23 +681,72 @@ class ConfigurationFile implements IConfigurationFile
      * This method updates the configuration file to include a new ICS subscription key.
      * If the key already exists, it does nothing.
      *
+     * @param string|null $configFilePath Path to rewrite; defaults to the live config file.
      * @return void
      */
-    public function EnableSubscription()
+    public function EnableSubscription(?string $configFilePath = null)
     {
-        $icsKey = $this->GetKey(ConfigKeys::ICS_SUBSCRIPTION_KEY);
-        if (!empty($icsKey)) {
+        $newKey = self::GenerateSubscriptionKeyIfNeeded(
+            $this->GetKey(ConfigKeys::ICS_ENABLED, new BooleanConverter()),
+            $this->GetKey(ConfigKeys::ICS_SUBSCRIPTION_KEY)
+        );
+
+        if ($newKey === null) {
             return;
         }
 
-        $configFile = ROOT_DIR . 'config/config.php';
+        $configFilePath ??= ROOT_DIR . 'config/config.php';
 
-        if (file_exists($configFile)) {
-            $newKey = '$conf[\'settings\'][\'ics\'][\'subscription.key\'] = \'' . BookedStringHelper::Random(20) . '\';';
-            $str = file_get_contents($configFile);
-            $str = str_replace('$conf[\'settings\'][\'ics\'][\'subscription.key\'] = \'\';', $newKey, $str);
-            file_put_contents($configFile, $str);
-            Configuration::SetInstance(null);
+        if (!is_readable($configFilePath)) {
+            return;
         }
+
+        $contents = file_get_contents($configFilePath);
+
+        if ($contents === false) {
+            return;
+        }
+
+        // Matches the current `'subscription.key' => '',` nested-array entry.
+        $updated = preg_replace(
+            '/([\'"]subscription\.key[\'"]\s*=>\s*)([\'"])\2/',
+            '${1}${2}' . $newKey . '${2}',
+            $contents,
+            1,
+            $count
+        );
+
+        if ($updated === null || $count === 0) {
+            return;
+        }
+
+        // Checking is_writable() first keeps the common read-only-config case from
+        // emitting a warning, but the return value still has to be checked because
+        // the write can fail for reasons the check cannot see.
+        if (!is_writable($configFilePath) || file_put_contents($configFilePath, $updated) === false) {
+            Log::Error('Could not write the generated ICS subscription key to %s', $configFilePath);
+            return;
+        }
+
+        Configuration::SetInstance(null);
+    }
+
+    /**
+     * Shared gate for ICS subscription key generation: returns a freshly generated
+     * key when ICS is enabled and no key is set yet, or null when nothing should change.
+     * Used both here (patching the live config file) and by ManageConfigurationPresenter
+     * (patching the in-memory settings array before it writes the config file).
+     *
+     * @param bool $icsEnabled
+     * @param string|null $existingKey
+     * @return string|null
+     */
+    public static function GenerateSubscriptionKeyIfNeeded(bool $icsEnabled, ?string $existingKey): ?string
+    {
+        if (!$icsEnabled || !empty($existingKey)) {
+            return null;
+        }
+
+        return BookedStringHelper::Random(32);
     }
 }
--- a/tests/Infrastructure/Config/EnableSubscriptionTest.php
+++ b/tests/Infrastructure/Config/EnableSubscriptionTest.php
@@ -1,0 +1,256 @@
+<?php
+
+declare(strict_types=1);
+
+require_once(ROOT_DIR . 'lib/Config/namespace.php');
+
+class EnableSubscriptionTest extends TestBase
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        Configuration::SetInstance(null);
+    }
+
+    public function testGeneratesKeyWhenIcsEnabledAndKeyEmptyInCurrentNestedFormat(): void
+    {
+        $path = $this->writeTempConfig(
+            <<<'PHP'
+<?php
+return [
+    'settings' => [
+        'ics' => [
+            'enabled' => true,
+            'subscription.key' => '',
+        ],
+    ],
+];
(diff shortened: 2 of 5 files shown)


lib/Config/Configuration.php
```diff
@@ -679,23 +680,58 @@ public function GetAdminEmail()
...
+            '${1}${2}' . $newKey . '${2}',
+            $contents,
+            1,
+            $count
+        );
+
+        if ($count > 0) {
+            file_put_contents($configFilePath, $updated);
             Configuration::SetInstance(null);
         }
```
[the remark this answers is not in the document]
lucs7: As the feature is not essential, it will not be backported for the old config style

lib/Config/Configuration.php
```diff
@@ -262,11 +262,12 @@ public function GetAdminEmail()
 
     /**
      * Enables ICS subscription with a random key if no key is set.
+     * @param string|null $configFilePath Path to rewrite; defaults to the live config file.
      * @return void
      */
-    public function EnableSubscription()
+    public function EnableSubscription($configFilePath = null)
```
JohnVillalovos, line 269: Can this use type-hints please?

lucs7:
# Codex Review of `73b180596` — ICS subscription key generation

## Recommendation

Merge upstream. No merge-blocking issues were found. This recommendation assumes that support for updating legacy `$conf[...]` configuration files is intentionally out of scope.

## Findings

### Low: Environment-controlled subscription keys can gain an unused fallback

`ManageConfigurationPresenter::EnableIcsSubscriptionIfNeeded()` operates on the file-backed settings after `Configurator::BuildConfig()` has preserved values controlled by environment variables. If `LB_ICS_SUBSCRIPTION_KEY` is set while the corresponding config-file value is empty, the presenter generates and writes a random fallback key anyway.

The environment value remains authoritative, so this does not break the active subscription key. However, it bypasses the configurator's environment-preservation behavior and changes the fallback value that becomes active if the environment variable is later removed. Consider skipping generation when `ConfigKeys::hasEnv(ConfigKeys::ICS_SUBSCRIPTION_KEY)` is true.

### Low: The unreadable-file path emits a PHP warning

`file_get_contents()` emits an `E_WARNING` before returning `false`. The new false-return guard prevents invalid content from being processed, but `EnableSubscriptionTest::testDoesNothingWhenConfigFileIsUnreadable()` consequently reports a PHPUnit warning rather than completing cleanly.

Consider handling the expected read failure without leaking a raw PHP warning, and logging it deliberately if operational visibility is desired.

## Commit message

The message follows the conventional commit structure, stays within the header-length limit, and includes the required `Assisted-by` trailer. `fix(ics)` would describe the primary purpose more precisely than `feat(ics)`, although the added automatic generation behavior makes the current type defensible. Capitalizing “ICS” and the sentence beginning with “this” would improve presentation.

lucs7:
# Claude Review: `73b180596` — `feat(ics): fix ICS subscription key generation if empty`

Repo state confirmed: single commit on `feature/ics-rework/ics-secret`, based directly
on `origin/develop` (`d4fbb9000`), no merge commits. Working tree clean except
unrelated untracked scratch files.

## Accepted by the maintainer: legacy config format is not supported

The new regex requires `=>`, so it does not match the legacy
`$conf['settings']['ics']['subscription.key'] = '';` form that the old `str_replace`
handled (I confirmed: legacy input yields `count=0`). `Configuration::Register()` still
loads that format with a deprecation notice, so legacy-format installs no longer get
automatic key generation from the resource/schedule/user toggle path.

**This is an accepted, intentional decision** — the maintainer has confirmed legacy
config support is not required here. Worth one line in the commit message so the
narrowed scope is a recorded choice rather than something a future bisector has to
rediscover; those installs will need to migrate to the new format (or set
`ics.subscription.key` by hand) to get a key.

## Findings

### 1. The new guard emits a PHP warning it's trying to suppress

`lib/Config/Configuration.php:703`

`file_get_contents()` raises `E_WARNING` *before* returning `false`, so the "treat a
read failure as a no-op" guard still spams the error log. The new test surfaces this:

```text
1 test triggered 1 PHP warning:
1) lib/Config/Configuration.php:703
file_get_contents(/tmp/enable-subscription-test-…): Failed to open stream: Permission denied
```

Full-suite warnings go 9 → 10. CI won't fail (`phpunit.yml` only greps for
`Tests:|OK \(`), but the fix is one character: `@file_get_contents(...)`, or gate on
`is_readable()`.

### 2. Interface signature not updated

`lib/Config/Configuration.php:69`

`IConfigurationFile::EnableSubscription()` still declares no parameters, and its
docblock still lacks `@param`. PHP permits implementations to add optional params and
PHPStan level 2 won't flag it, but the interface is now out of sync with all four
implementations (`Configuration`, `ConfigurationFile`, `FakeConfig`,
`FakeConfigFile`). Should be `EnableSubscription($configFilePath = null);`.

### 3. Two real behavior changes missing from the commit message

- **Key length 20 → 32.** `Random(20)` = 10 random bytes (80 bits); `Random(32)` = 16
  bytes (128 bits). A good security improvement, but it's silent in both the message
  and the diff context. It deserves a line — reviewers shouldn't have to
  reverse-engineer entropy changes to a secret.
- **`EnableSubscription()` is now gated on `ics.enabled`.** Previously it generated a
  key unconditionally; now the `GenerateSubscriptionKeyIfNeeded` gate short-circuits
  when ICS is off. This changes the three existing callers
  (`ManageResourcesPresenter:531`, `PersonalCalendarPresenter:41`,
  `ManageSchedulesPresenter:184`). It's defensible — a key is useless when feeds 404 —
  but it's a behavior change, not a refactor, and isn't mentioned.

### 4. Commit message style (minor)

`feat(ics): fix ICS subscription key generation if empty` claims `feat` while the
subject says "fix". `feat` is defensible since autogeneration-on-enable is new, but the
header is self-contradictory — consider
`feat(ics): generate subscription key when ICS is enabled`. The body also uses first
person ("In addition I added") and lowercase sentence starts, unlike the repo's usual
imperative style. No `Closes:` footer — is there an issue to reference?

### 5. Type declarations (minor)

CLAUDE.md asks for type declarations on new code.
`GenerateSubscriptionKeyIfNeeded($icsEnabled, $existingKey): ?string` has a return type
but no param types, and `$configFilePath = null` is untyped. Matching the untyped
legacy style of the surrounding file is a reasonable call for the
interface-constrained param, but the new static helper has no such constraint and could
be `(bool $icsEnabled, ?string $existingKey)`.

### 6. Small inconsistency between the two code paths

The presenter uses `$mergedSettings['ics']['enabled'] ?? ConfigKeys::ICS_ENABLED['default']`.
`??` only fires on null/unset, so an empty-string value gives `filter_var('')` →
`false`, whereas `GetKey()` on the file path would return the `true` default. Unlikely
to bite in practice, but the two paths disagree on that edge.

### 7. Test coverage gap (minor)

No presenter test for `ics.enabled` being absent from `mergedSettings` — i.e. the
default-`true` fallback branch, which is the one that differs from the file path per
finding 6.

Test placement is fine, incidentally: `tests/Infrastructure/Config/` already contains
topic-named files (`DeprecatedConfigKeysTest`, `ConfigKeyDefinitionsTest`), so
`EnableSubscriptionTest` matches local practice even though CLAUDE.md's rule literally
implies `ConfigurationTest`. The `posix_getuid()` root guard on the permissions test is
a nice touch and will work under GitHub Actions (`runner`, non-root).

## Should this be merged upstream?

**Yes — it fixes a genuine, user-visible bug, and the code is clean, tested, and passes
every gate.** The regex is correctly anchored to the key name with a quote
backreference, the shared `GenerateSubscriptionKeyIfNeeded` gate removes real
duplication, and closing the "admin enables ICS but the key stays empty until an
unrelated toggle" hole is the right call.

With legacy config support accepted as out of scope, nothing here blocks the merge. Two
one-line cleanups are worth doing first:

- **Finding 1** — one-character fix, and the guard doesn't do what the commit message
  says it does without it.
- **Finding 2** — one-line interface update.

And **finding 3** — expand the commit message; the entropy bump and the `ics.enabled`
gate are both things a future bisector needs to see, alongside the note that the legacy
config format is deliberately not handled. The rest are nits.

tests/Infrastructure/Config/EnableSubscriptionTest.php
```diff
@@ -0,0 +1,173 @@
+<?php
```
JohnVillalovos, line 1: What do you think about having the test make a copy of `config/config.dist.php` and doing a test against that?

That way if the config file changes in the future it will be caught.

lib/Config/Configuration.php
JohnVillalovos: Can keep the docblock also as it has a comment on the purpose of the variable but still want the type-hint please.

tests/Infrastructure/Config/EnableSubscriptionTest.php
lucs7: done

lucs7:
all resolved

JohnVillalovos (approved):
LGTM

Thanks @lucs7

PR #1656, opened 2026-08-22
fix(email): clear attachments before each send to stop them leaking

lucs7:
ServiceLocator::GetEmailService(): PHPMailer  Send() already clears recipients and reply-tos before each message but never called clearAttachments(), so PHPMailer's internal attachment list persisted across sends. A reservation-creation notification burst (owner, admins, participants, invitees, guests) sends several emails through that same instance, so each subsequent email inherited every prior message's attachment on top of its own.

Add a generic FakeEmailMessage and EmailServiceTest proving EmailService::Send() clears PHPMailer's attachment list before each message, so a second Send() on the same instance doesn't inherit the first message's attachment.

Assisted-by: Claude:claude-sonnet-4-6

### DIFF
--- a/lib/Email/EmailService.php
+++ b/lib/Email/EmailService.php
@@ -45,6 +45,7 @@ class EmailService implements IEmailService
     {
         $this->phpMailer->clearAllRecipients();
         $this->phpMailer->clearReplyTos();
+        $this->phpMailer->clearAttachments();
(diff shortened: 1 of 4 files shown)


JohnVillalovos (approved):
LGTM

Thanks @lucs7

PR #1657, opened 2026-08-22
Feature/ics rework/mime

lucs7:
This pull request enhances the email attachment handling system by allowing MIME types to be specified for string attachments, ensuring correct content types are used when sending emails. 

Depends on #1656


lucs7:
### Code review

Found 1 issue:

1. `ReservationEmailMessage::PopulateIcsAttachment()` hardcodes the ICS attachment's Content-Type as `method=REQUEST` for every reservation email, including cancellations and shares. `ReservationDeletedEmail` (a cancellation notice) and `ReservationShareEmail` (a plain share/export) don't override this method, so they'll also ship `method=REQUEST` in the attachment header. Per RFC 5546/iTip, a cancellation should use `method=CANCEL` and a share/export should use `method=PUBLISH` — with the wrong value, RFC 5546-aware clients (e.g. Gmail, Outlook) may render incorrect or missing Accept/Decline/Cancel scheduling UI for these emails.

https://github.com/LibreBooking/librebooking/blob/1e62cc7b4739c4087af1f0a4ce7496380c95daab/lib/Email/Messages/ReservationEmailMessage.php#L220-L227

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

lucs7:
Code review is true but mimics the current behaviour of the ics implementation. actual correct method follows in the next PR

JohnVillalovos (approved):
LGTM

Thanks @lucs7 

Sorry for the delay.

PR #1687, opened 2026-09-04
docs: fix three dead links in the install and developer docs

rawsun007:
Three links in the published docs return 404.

`docs/source/DEVELOPER-README.rst`

- PHPUnit: `phpunit.readthedocs.io/en/latest/writing-tests-for-phpunit.html` → `docs.phpunit.de/en/13.3/writing-tests-for-phpunit.html`. PHPUnit moved off Read the Docs.
- phpDocumentor: `docs.phpdoc.org/latest/guide/guides/running-phpdocumentor.html` → `docs.phpdoc.org/guide/getting-started/generating-documentation.html`. The page was renamed; the new one is the same "how to run it" content, and it is the entry their current nav links.

`docs/source/INSTALLATION.rst`

- XAMPP: `apachefriends.org/en/index.html` → `apachefriends.org/`. The `/en/index.html` path is gone; the root serves the downloads.

Each replacement returns 200; each old URL returns 404.

## What I deliberately left alone

- **WampServer**, on the line right after XAMPP. `wampserver.com/en/` 404s, but so does `wampserver.com/`, so their site is down or blocking rather than restructured, and I would rather not repoint a link based on an outage.
- `CHANGELOG.md` references `bookedscheduler.com/phpscheduleit`, also dead. That is the historical changelog, so rewriting old entries would falsify the record.
- The XAMPP link text reads `XAMMP` (transposed). I left it, since this PR is scoped to URLs — say the word and I will fix the spelling too.

Per the AI-assistance rule in `CONTRIBUTING.md`, both commits carry `Assisted-by: Claude Code:claude-opus-5`. Every URL claim above was verified with a request rather than inferred. Based against `develop`.

### DIFF
--- a/docs/source/DEVELOPER-README.rst
+++ b/docs/source/DEVELOPER-README.rst
@@ -80,15 +80,15 @@ Run ``composer install`` to install Composer-managed tools into
 ``composer build`` Builds / Packages a distributable relase inside
 ``/build`` configured via ``/build.xml``.
 
-`PHPUnit <https://phpunit.readthedocs.io/en/latest/writing-tests-for-phpunit.html>`__
+`PHPUnit <https://docs.phpunit.de/en/13.3/writing-tests-for-phpunit.html>`__
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 All classes should have good unit test coverage. The level of coverage
 is up to the developer and should be done when the code is sufficiently
 complex. Tests must all succeed for a final release.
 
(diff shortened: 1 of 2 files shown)


rawsun007:
Sorry for the noise — the first push failed three checks and all three were my fault. Now green (17/17).

- `sphinx` and `doc8` both flagged `DEVELOPER-README.rst:91: Title underline too short`. Those link lines are section titles, and my longer phpDocumentor URL outgrew the `~~~` underline beneath it. Underline extended to match the title exactly.
- `lint-commits` flagged `body-max-line-length` on both commits; the bodies are now wrapped at 72.

While fixing the commit messages I reset the branch to base, which made GitHub auto-close this PR since head briefly equalled base. That is why it shows closed and reopened — the diff is unchanged in intent: three URLs plus the underline.

PR #1691, opened 2026-09-05
fix: group CSV import ignoring read-only permissions on update

### LINKED ISSUE #1665: [Bug] Group CSV import silently ignores "Read Only Permissions" when updating an existing group
### Acknowledgements

## Summary

When importing groups from CSV via **Manage Groups → Import** with the "update existing" option enabled, the read-only resource permissions column is silently ignored if the group already exists. Full-access permissions update correctly; only the read-only branch is affected.

## Root cause

In `Presenters/Admin/ManageGroupsPresenter.php`, the `Import()` method has two branches: one that updates an existing group, and one that creates a new group.

The **update** branch (existing group) reads a property that doesn't exist on `GroupImportCsvRow`:

```php
// Presenters/Admin/ManageGroupsPresenter.php
if (!empty($row->permissionsView)) {
    $resourceIds = $this->GetImportResources($row->permissionsRead, $resourcesIndexed);
    $group->ChangeViewPermissions($resourceIds);
} 
```

### To Reproduce

1. Create a group named TestGroup (via the UI, or import it once).
2. Prepare a CSV with a Read Only Permissions column listing one or more resource names, for the row TestGroup
3. Go to Manage Groups → Import, upload the CSV, check "update existing groups" (or whatever the update-on-import option is labeled), and import.
4. Check TestGroup's resource permissions.

### Expected behavior

TestGroup should have view-only permissions on the resources listed in the CSV's Read Only Permissions column.

### Screenshots

_No response_

### Additional context

_No response_

### LibreBooking version

v5.3.0-9-gccaa6116a

santifpinformatica:
GroupImportCsvRow only defines permissionsRead, but the update
branch of ManageGroupsPresenter::Import() checked the nonexistent
property $row->permissionsView. That check always evaluated as
empty, so read-only permissions from the CSV were silently
skipped whenever an existing group was updated. Full-access
permissions were unaffected, since that branch checked the
correct property.

Fixed the condition to check $row->permissionsRead instead.

Fixes #1665

### DIFF
--- a/Presenters/Admin/ManageGroupsPresenter.php
+++ b/Presenters/Admin/ManageGroupsPresenter.php
@@ -605,7 +605,7 @@ class ManageGroupsPresenter extends ActionPresenter
                         $group->ChangeAllowedPermissions($resourceIds);
                     }
 
-                    if (!empty($row->permissionsView)) {
+                    if (!empty($row->permissionsRead)) {
                         $resourceIds = $this->GetImportResources($row->permissionsRead, $resourcesIndexed);
 
                         $group->ChangeViewPermissions($resourceIds);


santifpinformatica (commented):
fix: group CSV import ignoring read-only permissions on update #1691

PR #1708, opened 2026-09-14
style(lang): update ee_ee.php

qbicdesign:
Extends en_us now rather than en_gb due to locale differences. Current strings from en_us.php imported and translated


qbicdesign:
Specifically:

- date separator in estonian language is officially . not / hence the change of extends to en_us
- The language file was really out of date, I re-translated from scratch, but skipped the` // install` section so that it will fall back to en_us for that

